PR #13, opened 2023-05-19
Add metadata management

FelixTJDietrich:
## Motivation
Meta field was empty for suggest feedback and I tried populating it in the module's submissions_consumer but it did not propagate. 

## Description
Filled meta fields for exercises, submissions, and feedbacks and automatically load it. 
Also store feedback suggestions and automatically assign an id to them.

## Testing using playground
1. Send programming submissions (to the example module)
2. Look at printed metadata from module
3. Select submission and suggest feedback for it
4. Check that printed metadata is the one from before

Additionally, you can inspect the database via DBBeaver and check if the feedback suggestions gets stored.

## Issues
- Feedback ids, maybe we should have a separate table for feedback suggestions, otherwise they might collide with the ones the LMS sends us.


pal03377 (commented):
Looks good! I have 2 minor points we should discuss before merging

pal03377 (approved):
Very good! I just removed 3 unused imports

PR #15, opened 2023-05-20
Playground auth secret input

pal03377:
The branch builds on #12

Add an input field for the API secret to the playground (only required for sending requests to production setups).

screenshot-2023-05-20_001368


FelixTJDietrich (approved):
I get an unauthenticated if I do it like this, but I might be doing it wrongly:

Screenshot 2023-05-23 at 11 12 55
Screenshot 2023-05-23 at 11 13 07

One thing that is strange to me is, that the log message I added appears in the `assessment_module_manager`, shouldn't it appear in the module? Doesn't the module check the secret from the assessment module manager? I'm confused.

Other than that the code looks good to me, and if it works for you in production mode we can merge it 👍

PR #18, opened 2023-05-20
Playground refactoring: Separate data from code

pal03377:
This PR builds on #15 

This is a refactoring of the playground: Now the data and the code are cleanly separated. Also, there is a new option to send multiple feedbacks at once via the playground for development convenience.


FelixTJDietrich (approved):
LGTM and works as expected. In the future we might need a way to collapse the responses, probably will get to much.

PR #17, opened 2023-05-20
Add module programming llm

FelixTJDietrich:
## Description
- Adds the first iteration of the programming llm module
- Adds complete sample programming exercise to playground for testing
  - Sample solution, tests, and template + grading instructions and problem statement
- Fix async types

## Issues
- Returned feedback ids are null

PR #20, opened 2023-05-21
Playground hosting

pal03377:
This PR builds on the changes to the playground from #17 and enables hosting it on a server as well.

This will enable testing programming exercise modules on deployed instances, because then they can fetch the programming repositories from an actually reachable playground URL.


FelixTJDietrich (approved):
Other than the comments the code LGTM. I tested it locally without docker and the requests go to the right location.

FelixTJDietrich (approved):
Other than the comments the code LGTM. I tested it locally without docker and the requests go to the right location.

PR #21, opened 2023-05-21
Enable mypy checks

pal03377:
This PR enables mypy, which -as I noticed - apparently wasn't enabled before.

I'll also fix type issues on the current `develop` branch here that arise because of this.


FelixTJDietrich (approved):
LGTM, thanks for adding this :)

PR #26, opened 2023-06-09
CoFee Submission Selection

pal03377:
This PR implements the submission selection for the CoFee module.

I studied the `TextAssessmentQueueService.java` (and related files) in Artemis to find out how exactly the next submission to grade is determined. It's a calculation of information gain by submission and then choosing the highest one. I tried to replicate & simplify the information gain calculation and comparison directly. The resulting code is much simpler than the one to be found in Artemis, but I'm sure it resembles the same logic.


FelixTJDietrich (approved):
I still could not test it locally because of Athena not running but the code looks good to me

FelixTJDietrich (approved):
I still could not test it locally because of Athena not running but the code looks good to me

PR #30, opened 2023-06-11
ThemisML Module

### LINKED ISSUE #22: Referenced Feedback Schema
## Motivation / Description

We currently have the classes `Feedback`, `ProgrammingFeedback(Feedback)`, and `TextFeedback(Feedback)` and Feedback contains a `reference` attribute:
```
reference: Optional[str] = Field(description="A optional reference to the location in the submission where the "
                                                 "feedback applies.",
                                     example="file:src/pe1/MergeSort.java_line:12")
```

Maybe we should move the reference to its descendants and adjust it for the exercise type:
- Text Exercise
    - `unreferenced`
    - `referenced` to some `TextBlock`
- Programming Exercise
    - `unreferenced`
    - `unreferenced` but referencing a `file_path`
    - `referenced` referencing a `file_path` and a line, line range, or span of code
    - `suggestion` referencing a span of code and a potential replacement
        - Maybe we have multiple suggestions in a ProgrammingFeedback

We can store it like this somehow in the database and compute for example a formatted `text` (e.g. `File src/pe1/MergeSort.java at line 12`) for the LMS to use.

Maybe we can agree on a schema changes here and also on the model type changes.

pal03377:
This PR introduces a new ThemisML module based on https://github.com/ls1intum/Themis-ML

Currently, it's IMO not fast enough for production usage, but the suggestion algorithm already works for whole methods.

To test: 
- Send submissions of the "Hello World" exercise
- Send feedback of the "Hello World" exercise
- Request feedback suggestions for the submission with ID 104, which references the same code as submission 101 with feedback (a bit simplistic, I know - but works for basic testing)
- You should get back a suggestion:
screenshot-2023-06-11_001427

I also changed the feedback schemas and models in a way that could already solve #22 in the process (closes #22)


FelixTJDietrich:
On send feedback of submission `101` and feedback `101` I get:
Code `422`
```
{
  "module_name": "module_programming_themisml",
  "status": 422,
  "data": {
    "detail": [
      {
        "loc": [
          "body",
          "feedback",
          "exercise_id"
        ],
        "msg": "field required",
        "type": "value_error.missing"
      },
      {
        "loc": [
          "body",
          "feedback",
          "submission_id"
        ],
        "msg": "field required",
        "type": "value_error.missing"
      },
      {
        "loc": [
          "body",
          "feedback",
          "detail_text"
        ],
        "msg": "field required",
        "type": "value_error.missing"
      },
      {
        "loc": [
          "body",
          "feedback",
          "text"
        ],
        "msg": "field required",
        "type": "value_error.missing"
      },
      {
        "loc": [
          "body",
          "feedback",
          "credits"
        ],
        "msg": "field required",
        "type": "value_error.missing"
      },
      {
        "loc": [
          "body",
          "feedback",
          "meta"
        ],
        "msg": "field required",
        "type": "value_error.missing"
      }
    ]
  }
}
```

Am I doing something wrong?

athena/athena/models/db_programming_feedback.py
```diff
@@ -1,15 +1,38 @@
...
+from .db_programming_submission import DBProgrammingSubmission
 from .db_feedback import DBFeedback
 
 
 class DBProgrammingFeedback(DBFeedback, Base):
     __tablename__ = "programming_feedbacks"
 
+    file_path: str = Column(String)  # type: ignore
+    line_start: int = Column(Integer)  # type: ignore
+    line_end: int = Column(Integer)  # type: ignore
```
FelixTJDietrich: I think the type for all three should be `Optional[...]`

athena/athena/schemas/feedback.py
```diff
@@ -12,16 +12,34 @@ class Feedback(Schema, ABC):
...
+
+    def __init__(
+            self,
+            id: Optional[int] = None,
+            exercise_id: int = -1,
+            submission_id: int = -1,
+            detail_text: str = "",
+            text: str = "",
+            credits: float = 0.0,
+            meta: dict = None,  # type: ignore # This is None to avoid mutable default arguments
```
FelixTJDietrich: Wouldn't `Optional[dict]` still make sense? You are removing the optional with `or {}` below

athena/athena/models/db_programming_feedback.py
```diff
@@ -1,15 +1,38 @@
...
+
     exercise_id = Column(Integer, ForeignKey("programming_exercises.id", ondelete="CASCADE"), index=True)
     submission_id = Column(Integer, ForeignKey("programming_submissions.id", ondelete="CASCADE"), index=True)
 
     exercise = relationship("DBProgrammingExercise", back_populates="feedbacks")
     submission = relationship("DBProgrammingSubmission", back_populates="feedbacks")
+
+
+    def get_code(self):
```
FelixTJDietrich: Maybe add a return type and mention in the method description that it can throw a `ValueError`. Maybe it is also a bit confusing, maybe call it `get_referenced_code`? Also, I don't know if we allow it but if we have only `file_path` set without `line_start` and `line_end`, I would assume that it gives me the code for the whole file.

athena/athena/models/db_programming_exercise.py
```diff
@@ -8,10 +8,10 @@
...
     __tablename__ = "programming_exercises"
 
-    programming_language = Column(String)
-    solution_repository_url = Column(String)
-    template_repository_url = Column(String)
-    tests_repository_url = Column(String)
+    programming_language: str = Column(String)  # type: ignore
+    solution_repository_url: str = Column(String) # type: ignore
+    template_repository_url: str = Column(String) # type: ignore
+    tests_repository_url: str = Column(String)  # type: ignore
```
FelixTJDietrich: I'm a bit confused by the `nullable` in `sqlalchemy`, I think by default everythin is nullable now? Shouldn't we then have `nullable=False` for most db columns? At least here

athena/athena/schemas/programming_feedback.py
```diff
@@ -1,27 +1,48 @@
...
 
 from .feedback import Feedback
-from ..helpers.programming import format_feedback_text, format_feedback_reference
+
 
 class ProgrammingFeedback(Feedback):
     """Feedback on a programming exercise."""
+    file_path: Optional[str] = Field(None, example="src/pe1/MergeSort.java")
+    line_start: Optional[int] = Field(None, example=1)
+    line_end: Optional[int] = Field(None, example=2)
```
FelixTJDietrich, line 14: Maybe we can have a more safer type for this. For example right now I could set only `line_end` to `None`, but `line_start` and `line_end` should only be settable when `file_path` is set and then both have to be not `None`.

I guess you mention below that single line feedback is when `line_end` is `None` 🤔, but maybe that is also confusing.

athena/athena/schemas/programming_feedback.py
```diff
@@ -1,27 +1,48 @@
...
+        self.file_path = file_path
+        self.line_start = line_start
+        self.line_end = line_end
+
+
+    @property
+    def line_range(self) -> Tuple[int, int]:
+        """Return the line range of this feedback, even if it is only a single line (line_end is None)."""
+        if self.line_start is None:
+            raise ValueError("Feedback does not have a line start")
```
FelixTJDietrich: Maybe mention that it throws an `ValueError` in the docstring or think about changing the return type.

athena/athena/schemas/programming_submission.py
```diff
@@ -18,4 +19,13 @@ def get_zip(self) -> ZipFile:
...
     def get_repository(self) -> Repo:
         """Return the submission repository as a Repo object."""
-        return get_repository(self.repository_url)
\ No newline at end of file
+        return get_repository(self.repository_url)
+    
+    def get_code(self, file_path: str):
+        """
+        Fetches the code from the submission repository.
+        Might be quite an expensive operation! If you need to fetch multiple files, consider using get_zip() instead.
```
FelixTJDietrich, line 27: I guess some kind of caching would be nice, but we can do it as a follow up issue and discuss how we should handle the repository downloads in general.

FelixTJDietrich: Alternativly, there is already some caching in `get_repository`. You can use that and then return the file's content from there.

athena/athena/schemas/programming_submission.py
```diff
@@ -18,4 +19,13 @@ def get_zip(self) -> ZipFile:
 
     def get_repository(self) -> Repo:
         """Return the submission repository as a Repo object."""
-        return get_repository(self.repository_url)
\ No newline at end of file
+        return get_repository(self.repository_url)
+    
+    def get_code(self, file_path: str):
```
FelixTJDietrich: A return type would be nice

athena/athena/schemas/text_feedback.py
```diff
@@ -1,18 +1,34 @@
...
 class TextFeedback(Feedback):
     """Feedback on a text exercise."""
-    def get_start_index(self) -> int:
-        """Helper function to get the start index of the feedback from its reference."""
-        # reference is a string of the form "start_index-end_index"
-        if self.reference is None:
-            raise ValueError("Feedback does not have a reference.")
-        return int(self.reference.split('-')[0])
+    index_start: Optional[int] = Field(description="The start index of the feedback in the submission text.", example=0)
+    index_end: Optional[int] = Field(description="The end index of the feedback in the submission text.", example=10)
```
FelixTJDietrich: Here we could also have a safer type, at least for the schema class, I think for the db it is fine. What do you think?

athena/athena/schemas/text_feedback.py
```diff
@@ -1,18 +1,34 @@
...
-        return int(self.reference.split('-')[1])
+    def __init__(
+            self,
+            id: Optional[int] = None,
+            exercise_id: int = -1,
+            submission_id: int = -1,
+            detail_text: str = "",
+            text: str = "",
+            credits: float = 0.0,
+            meta: dict = None,  # type: ignore # This is None to avoid mutable default arguments
```
FelixTJDietrich: Maybe still use `Optional` and do the `or {}` as you did with programming feedback.

module_programming_llm/module_programming_llm/basic/basic_feedback_provider.py
```diff
@@ -38,12 +38,12 @@ async def suggest_feedback(exercise: Exercise, submission: Submission) -> List[F
         if submission_content is None:
             continue
             
-        problem_statement = exercise.meta['file_problem_statements'][file_path]
+        problem_statement = exercise.meta['file_problem_statements'].get(file_path)
         if problem_statement is None:
```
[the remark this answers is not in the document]
FelixTJDietrich: Yeah you are right, sometimes where is an issue and the `file_problem_statements` is not present

module_programming_llm/module_programming_llm/basic/basic_feedback_provider.py
```diff
@@ -70,6 +70,8 @@ async def suggest_feedback(exercise: Exercise, submission: Submission) -> List[F
 
     # Completion
     chain = LLMChain(llm=chat, prompt=chat_prompt)
+    if not input_list:
+        return []
```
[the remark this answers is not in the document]
FelixTJDietrich: 🙏

module_text_cofee/module_text_cofee/models/db_text_block.py
```diff
@@ -8,11 +8,11 @@
...
-    id = Column(String, primary_key=True, index=True)
-    text = Column(String)
-    start_index = Column(Integer)
-    end_index = Column(Integer)
-    added_distance = Column(Float, nullable=True, default=None)
+    id: str = Column(String, primary_key=True, index=True)  # type: ignore
+    text: str = Column(String)  # type: ignore
+    index_start: int = Column(Integer)  # type: ignore
+    index_end: int = Column(Integer)  # type: ignore
+    added_distance: float = Column(Float, nullable=True, default=None)  # type: ignore
```
FelixTJDietrich: The confusion continues ^^ is `sqlalchemy` now by default nullable or not? I think we have to go over all columns and check this again.

module_text_llm/module_text_llm/helpers/utils.py
```diff
@@ -35,7 +35,7 @@ def get_sentence_spans(content: str) -> List[Tuple[int, int]]:
     return sentence_spans
 
 
-def parse_line_number_reference_as_span(reference: str, content: str) -> Optional[str]:
+def parse_line_number_reference_as_span(reference: str, content: str) -> Tuple[Optional[int], Optional[int]]:
```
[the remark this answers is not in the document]
FelixTJDietrich: True, I guess I will change it in the future depending how the final prompt reference structure is.

playground/examples/exercise-1/submissions/102/BinarySearch.java
```diff
@@ -6,7 +6,7 @@ public class BinarySearch implements SearchStrategy {
     @Override
     public int performSearch(List<Chapter> chapters, String chapterName) {
         int low = 0;
-        int high = chapters.size() - 1;
+        int high = chapters.size() - 1
```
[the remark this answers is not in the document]
FelixTJDietrich: 👍

playground/src/model/feedback.ts
```diff
@@ -2,9 +2,10 @@ type Feedback = {
   id: number;
   exercise_id: number;
   submission_id: number;
-  detail_text: string;
-  reference?: string;
   text: string;
+  detail_text: string;
+  index_start: number;
+  index_end: number;
```
FelixTJDietrich, line 8: I guess here we could also use a safer type for the reference structure. Also right now the feedback cannot be unreferenced?

FelixTJDietrich (changes requested):
Some feedback already on everything except themis ^^ Will continue looking at it.

FelixTJDietrich (changes requested):
Some feedback already on everything except themis ^^ Will continue looking at it.

module_programming_themisml/module_programming_themisml/extract_methods/languages/JavaLexer.py
```diff
@@ -0,0 +1,648 @@
+# Generated from JavaLexer.g4 by ANTLR 4.11.1
```
FelixTJDietrich, line 1: Did you customize this? If not then it could maybe be a file that can be gitignored and generated on init etc.?

FelixTJDietrich:
Okay, I looked also at the ThemisML module now and the code looks generally good to me. Though I did not 100% get how it works in detail but I think that is also not too important. I just could not test it because of the error I posted above.

playground/src/model/feedback.ts
pal03377: I fixed that it now can be unreferenced. What's do you have in mind for the safer type?

athena/athena/schemas/programming_feedback.py
pal03377: I now added pydantic validators to ensure that
- either `line_start` and `line_end` are both None or they are both ints (`line_end` possibly is auto-populated in the validator)
- `line_start` is never larger than `line_end`

athena/athena/schemas/programming_feedback.py
pal03377: The helper function is now gone in favor of validators

athena/athena/schemas/feedback.py
pal03377: Resolved by just using pure pydantic models

athena/athena/models/db_programming_feedback.py
pal03377: Very valuable feedback! I've implemented it that way 👍

athena/athena/models/db_programming_exercise.py
pal03377: I've gone through all models and schemas to solve this for all of them. Now, all not-nullable columns should be marked as such.

athena/athena/schemas/programming_submission.py
pal03377: I suggest to do that as a follow-up

athena/athena/schemas/text_feedback.py
pal03377: I've added two Pydantic validators 👍

athena/athena/schemas/text_feedback.py
pal03377: outdated now, with pure pydantic schemas

module_text_cofee/module_text_cofee/models/db_text_block.py
pal03377: I did now! 🎉  It's nullable by default

module_programming_themisml/module_programming_themisml/extract_methods/languages/JavaLexer.py
pal03377: I thought of this, but then I would have to download the grammar files from https://github.com/antlr/grammars-v4/tree/master/java/java and compile them. This would be an extra setup step for local developers, which wouldn't bring that much value, on the other hand. What do you think? I'm still a bit unsure

pal03377 (commented):
Fixed the issue you mentioned, let me know if it works now. Also worked on all other feedbacks. Thanks already!

playground/src/model/feedback.ts
FelixTJDietrich: Nevermind let's just leave it like this. There seems to not be a good way to have an enum with associated values.

I thought about something like this Swift code (for programming exercises):
```
enum FeedbackType {
    case unreferenced
    case unreferencedFile(filePath: String)
    case referenced(filePath: String, lineRange: (start: Int, end: Int))
}
```

module_programming_themisml/module_programming_themisml/extract_methods/languages/JavaLexer.py
FelixTJDietrich: Yeah, I think it makes sense to commit it now. It just makes it harder to get started.

FelixTJDietrich (approved):
Code LGTM :)
I also tested it now with the playground and it runs for me at least. I could not find submission 104 in my dropdown but at least it responded successfully

PR #32, opened 2023-06-14
Refactor playground and allow for additional evaluation data

FelixTJDietrich:
## Motivation

I want to add some evaluation scripts for the modules in the future and need to load in the data somehow. The evaluation data is sensitive and also quite large, so I would not commit it. Additionally, the example data proved to be quite useful so we should keep it for debugging etc. 

The example data has its place and we should keep adding good examples but we also need a way to test on other data.

### Why in the playground?

I thought about adding a separate cli script that just runs an evaluation, but I think adding that functionality to the playground will be of more value. We can debug faster and also extend it with some data annotation features for human evaluation etc.

## Description

- Add option to choose between example data and evaluation data
- Refactor the api routes to support this
- General improvements/refactorings to the playground
- Simplify example data and move it to `data/example`
- Add a gitignored `data/evaluation` directory for evaluation data
- Run prettier (maybe on too many files 🤔)


pal03377 (approved):
I tested it, it works well! The code also looks good. I would like to have a VS Code extension recommendation for Prettier as well as a project-wide configuration to always prettify on save though -> #34

While testing, I noticed an issue with the exercise-1 test data, which had a `text` present. I've removed that with an additional commit right here.

Also, I couldn't test the evaluation data part myself but I probably will gain access to some data in the near future. Do you have some script or something similar to generate the evaluation data format for the playground here from the Artemis DB? In the future, it might also be really convenient to be able to directly plug the playground into the local Artemis DB 🤔

PR #33, opened 2023-06-15
Remove student ID

pal03377:
It's not needed for any module right now. I previously thought that ThemisML needed it, but I was wrong about that. Therefore I've removed the field.


FelixTJDietrich (approved):
Code LGTM sorry for late approval!

PR #43, opened 2023-07-04
Add example env file to ThemisML module

pal03377:
### Motivation and Context

In #30, I apparently forgot to add an example env file for the ThemisML module. This PR adds that.

### Description

Add `module_programming_themisml.env` file into the `env_example` module following the example of other such module files. There is no special configuration for ThemisML, so it's the same as the one for `module_example`.

### Steps for Testing

None

### Screenshots

-/-


pal03377:
Because the changes are trivial, I'm merging this right away.

PR #44, opened 2023-07-04
Add module config for `module_example` and `module_text_llm`

FelixTJDietrich:
### Motivation and Context

We cant to test out multiple module configurations mostly for evaluation purposes, but it can also be used by the LMS to have some more control over the module. For my llm modules I want to be able to configure multiple approaches, multiple LLMs, and also prompts for the LLMs.

Another thing is that we might want to configure the module to be in debug mode or evaluation mode and send back debug data or evaluation data. For example for the LLMs, we would like to track the token usage, or the exact prompts used, etc.

#### Small illustration
image

#### List of changes
- [x] Add `config_schema_provider` decorator to `athena` (GET `/config_schema` endpoint for module sending back possible config options)
- [x] Also proxy GET requests in `assessment_module_manager` to modules
- [x] Add example `config_schema_provider` to `module_example`
- [x] Fix error that made `module_example` not run
- [x] Playground updates
    - [x] Display `/config_schema` request in `Module Requests` as `Get Config Schema`
    - [x] Display configs using `react-jsonschema-forms` in the `BaseInfoHeader` (it tries to render the schema if it is found, config component can be customized)
    - [x] Send module config with request headers to Athena (if enabled)
- [x] Add example of how to consume the config in `module_example`
- [x] Update `module_llm_text` to support azure and non-azure credentials at the same time
- [x] Add `config_schema_provider` to `module_text_llm`
    - [x] Available models (LLMs)
    - [x] Available approaches 
    - [x] Available prompts and their plain text (using `monaco` editor)
- [x] Rework openai model config
- [x] Use config in `module_text_llm`
    - [x] Use model
    - [x] Use approach
    - [x] Use prompts
- [x] Add `emit_meta` and `get_meta` to `athena` and include the emitted metadata with the request response

### Steps for Testing

**Prerequisites:** 
Update environment variables for `module_text_llm` or alternatively use `module_example` for testing, but it is less cool.

1. Do all requests in the playground without overriding the config
2. Do all requests in the playground with overriding the config for `module_text_llm` in the BaseInfoHeader
    - Setting debug to true sends back metadata (prompt & result) in the request feedback suggestions request
    - Prompt from the metadata should be the same of the config

### Screenshots

localhost_3000_playground (2)

Screenshot cut off....

localhost_3000_playground (5)


FelixTJDietrich:
I don't know where the Prospector Linting error comes from. It says cyclic-import for `module_text_llm/prompts/suggest_feedback_basic.py` but that file does not import anything...

assessment_module_manager/assessment_module_manager/module/request_to_module.py
```diff
@@ -32,15 +32,24 @@ async def find_module_by_name(
     return None
 
 
-async def request_to_module(module: Module, path: str, data: dict) -> ModuleResponse:
+async def request_to_module(module: Module, module_config: Optional[str], path: str, data: Optional[dict], method: str = "POST") -> ModuleResponse:
```
pal03377: Do you need the default value for `method`? I think it would make more sense to omit it.

athena/athena/endpoints.py
```diff
@@ -23,7 +26,12 @@
 }
 
 
-def submissions_consumer(func: Union[Callable[[E, List[S]], None], Callable[[E, List[S]], Coroutine[Any, Any, None]]]):
+def submissions_consumer(func: Union[
+    Callable[[E, List[S]], None],
+    Callable[[E, List[S], Optional[dict]], None],
+    Callable[[E, List[S]], Coroutine[Any, Any, None]],
+    Callable[[E, List[S], Optional[dict]], Coroutine[Any, Any, None]]
```
pal03377: Looks relatively clean, considering how complex those type checks can get! Could you also add comments to these lines, including examples of how such functions would look like?

athena/athena/endpoints.py
```diff
@@ -35,27 +43,54 @@ def submissions_consumer(func: Union[Callable[[E, List[S]], None], Callable[[E,
...
-            submissions: List[submission_type]):
+            submissions: List[submission_type],
+            module_config: Optional[str] = Header(None, alias="X-Module-Config"),
+            metadata: Dict[str, Any] = Depends(get_meta)):
+        if module_config:
+            try:
+                module_config = json.loads(module_config)
+            except json.JSONDecodeError:
+                logger.warning("Invalid module config received: %s", module_config)
+                module_config = None
```
pal03377: Why not let the server return an error `400`? This way, the error would be much more discoverable.

athena/athena/endpoints.py
```diff
@@ -35,27 +43,54 @@ def submissions_consumer(func: Union[Callable[[E, List[S]], None], Callable[[E,
...
+                    submission_meta = get_stored_submission_meta(stored_submission) or {}
+                    submission_meta.update(stored_submission.meta)
+                    submissions_dict[stored_submission.id].meta = submission_meta
         submissions = list(submissions_dict.values())
 
+        store_exercise(exercise)
+        store_submissions(submissions)
+
+        args = [exercise, submissions]
+        if len(inspect.signature(func).parameters) > 2:
```
pal03377: IMO this check should be improved. Can you detect that there is a parameter named `module_config` instead? Then, you can pass that to the `kwargs` of the function.

athena/athena/endpoints.py
```diff
@@ -81,22 +125,31 @@ async def wrapper(
...
+            args.append(module_config)
 
         # Select the submission
         if inspect.iscoroutinefunction(func):
-            submission = await func(exercise, submissions)
+            submission = await func(*args)
         else:
-            submission = func(exercise, submissions)
+            submission = func(*args)
```
pal03377: You could create a helper function for this part of the code

athena/athena/endpoints.py
```diff
@@ -110,24 +163,43 @@ def feedback_consumer(func: Union[Callable[[E, S, F], None], Callable[[E, S, F],
...
-            feedback: feedback_type):
+            feedback: feedback_type,
+            module_config: Optional[str] = Header(None, alias="X-Module-Config"),
+            metadata: Dict[str, Any] = Depends(get_meta)):
+        if module_config:
+            try:
+                module_config = json.loads(module_config)
+            except json.JSONDecodeError:
+                logger.warning("Invalid module config received: %s", module_config)
+                module_config = None
```
pal03377: Can you also put this into a helper? Maybe you can even switch the `module_config = Header(...)` part in the parameters to another fastapi `Depends`... I'm not sure

athena/athena/metadata.py
```diff
@@ -0,0 +1,26 @@
+from fastapi import Request
```
pal03377, line 35: I'm missing some documentation on the top of this file about its purpose

athena/athena/metadata.py
```diff
@@ -0,0 +1,26 @@
...
+    async def dispatch(self, request: Request, call_next: RequestResponseEndpoint):
+        # Init metadata in context variable
+        metadata_context.set({})
+        response = await call_next(request)
+        return response
+
+
+def get_meta() -> Dict[str, Any]:
+    # This will return an empty dict if no metadata has been set
+    return metadata_context.get({})
```
pal03377: I like this solution; it's pretty cool! Also, it's still simple to understand how to use it as a module developer

athena/athena/metadata.py
```diff
@@ -0,0 +1,26 @@
...
+class MetaDataMiddleware(BaseHTTPMiddleware):
+    async def dispatch(self, request: Request, call_next: RequestResponseEndpoint):
+        # Init metadata in context variable
+        metadata_context.set({})
+        response = await call_next(request)
+        return response
+
+
+def get_meta() -> Dict[str, Any]:
+    # This will return an empty dict if no metadata has been set
```
pal03377: I think there should be a more comprehensive documentation of this function as a `"""` docstring. The short information here should be included there.

athena/athena/metadata.py
```diff
@@ -0,0 +1,26 @@
+from fastapi import Request
+from starlette.middleware.base import BaseHTTPMiddleware, RequestResponseEndpoint
+from typing import Dict, Any
+import contextvars
+
+
+metadata_context: contextvars.ContextVar[Dict[str, Any]] = contextvars.ContextVar("metadata")
+
+
+class MetaDataMiddleware(BaseHTTPMiddleware):
```
pal03377, line 42: A usage example (documentation) here would be good

athena/athena/metadata.py
```diff
@@ -0,0 +1,26 @@
...
+        return response
+
+
+def get_meta() -> Dict[str, Any]:
+    # This will return an empty dict if no metadata has been set
+    return metadata_context.get({})
+
+
+def emit_meta(key: str, value: Any):
+    # Get the current metadata, add the new key-value pair
```
pal03377: I'd like to see another docstring here.

docs/module/structure.rst
```diff
@@ -95,6 +95,40 @@ Example:
                 )
             ]
 
+Provide Config Schema
```
pal03377, line 97: Well done documenting it here!

env_example/module_text_llm.env
```diff
@@ -2,10 +2,20 @@ PRODUCTION=1
 SECRET=12345abcdef
```
pal03377, line 2: Please remind me to change the config on Athena TS3 to make this work after this PR is merged

module_example/module_example/__main__.py
```diff
@@ -72,5 +95,14 @@ def suggest_feedback(exercise: Exercise, submission: Submission) -> List[Feedbac
     ]
 
 
+# Optional: Provide configuration options for the module if needed
+@config_schema_provider
+def available_config_schema() -> dict:
+    # Custom configuration options
+    # Ideally return a schema RFC8927 compliant json schema or something similar
+    # pydantic with `.schema()` mostly does the job
+    return Configuration.schema()
```
pal03377: I think this position is correct. The configuration endpoint is listed on top in the playground, but I think it would be better suited on the bottom like this (although the actual configuration can remain on top in the playground, of course). Most importantly to me, the order of endpoints should be consistent everywhere.

pal03377: This `Configuration.schema` is so cool! Can we have a class decorator for the `Configuration` class? This way, you wouldn't even need to define the endpoint here, just tell `athena` which is the configuration pydantic schema to use!

module_text_llm/module_text_llm/__main__.py
```diff
@@ -26,9 +27,19 @@ def process_incoming_feedback(exercise: Exercise, submission: Submission, feedba
 
 
 @feedback_provider
-async def suggest_feedback(exercise: Exercise, submission: Submission) -> List[Feedback]:
+async def suggest_feedback(exercise: Exercise, submission: Submission, module_config: Optional[dict]) -> List[Feedback]:
     logger.info("suggest_feedback: Suggestions for submission %d of exercise %d were requested", submission.id, exercise.id)
-    return await suggest_feedback_basic(exercise, submission)
+
+    config: Configuration = default_config
+    if module_config is not None:
```
pal03377: Can we guarantee to module developers that the `module_config` will never be `None` if they request it? If none is provided, we could just generate it from the defaults given by the pydantic schema.

module_text_llm/module_text_llm/config.py
```diff
@@ -0,0 +1,38 @@
...
+from module_text_llm.helpers.models import model_config, default_model_config
+from .prompts.suggest_feedback_basic import system_template, human_template
+
+
+class Prompt(BaseModel):
+    system_message: str
+    human_message: str
+
+
+class BasicPrompt(Prompt):
```
pal03377: Simplify:
```suggestion
class BasicPrompt(BaseModel):
```

module_text_llm/module_text_llm/config.py
```diff
@@ -0,0 +1,38 @@
...
+
+default_config = Configuration(
+    approach=BasicApproachConfig(
+        model=default_model_config,
+        prompt=BasicPrompt(
+            system_message=system_template,
+            human_message=human_template
+        )
+    )
+)
```
pal03377: You should be able to do this with only pydantic defaults. Then, you don't need this variable, and you can let the `athena` package handle the given-value-or-default-choice part.

module_text_llm/module_text_llm/helpers/models/__init__.py
```diff
@@ -0,0 +1,36 @@
...
+model_config = OpenAIModelConfig
+
+# model names -> models
+available_models: dict[str, BaseLanguageModel] = {
+    **available_openai_models,
+}
+
+default_model_key = os.environ.get("LLM_DEFAULT_MODEL")
+if default_model_key is None or default_model_key not in available_models:
+    raise EnvironmentError(f"LLM_DEFAULT_MODEL is not set correctly, available models:\n{', '.join(available_models)}")
```
pal03377: Please describe how to set `LLM_DEFAULT_MODEL` correctly in this error message. (It would suffice to reference the corresponding example file.)

module_text_llm/module_text_llm/helpers/models/__init__.py
```diff
@@ -0,0 +1,36 @@
...
+model = available_models[default_model_key]
+
+default_model_config = OpenAIModelConfig(
+    model_name=OpenAIModel[default_model_key],
+    temperature=0,
+    max_tokens=1024,
+    top_p=1,
+    presence_penalty=0,
+    frequency_penalty=0,
+)
```
pal03377: This should also be possible with only pydantic defaults, I believe.

module_text_llm/module_text_llm/helpers/models/model_config.py
```diff
@@ -0,0 +1,10 @@
+from abc import ABC, abstractmethod
+from pydantic import BaseModel
+from langchain.base_language import BaseLanguageModel
+
+
+class ModelConfig(BaseModel, ABC):
+    
+    @abstractmethod
+    def get_model(self) -> BaseLanguageModel:
+        pass
```
pal03377, line 10: Why not have a `BaseLanguageModel` directly? What do you need this abstract factory class for?

module_text_llm/module_text_llm/helpers/models/openai.py
```diff
@@ -0,0 +1,312 @@
...
+
+def _use_openai_credentials():
+    openai.api_type = "open_ai"
+    openai.api_key = os.environ.get("LLM_OPENAI_API_KEY")
+    openai.api_base = "https://api.openai.com/v1"
+    openai.api_version = None
+
+
+openai_available = os.environ.get("LLM_OPENAI_API_KEY") is not None
+azure_openai_available = os.environ.get("LLM_AZURE_OPENAI_API_KEY") is not None
```
pal03377: Oof, that's quite ugly. I remember the chats with you about the pain to get this to work, however. Is there some planned point in the future where we will remove this again?

module_text_llm/module_text_llm/helpers/models/openai.py
```diff
@@ -0,0 +1,312 @@
+# mypy: ignore-errors
```
pal03377: Why?

module_text_llm/module_text_llm/helpers/models/openai.py
```diff
@@ -0,0 +1,312 @@
...
+    openai.api_version = None
+
+
+openai_available = os.environ.get("LLM_OPENAI_API_KEY") is not None
+azure_openai_available = os.environ.get("LLM_AZURE_OPENAI_API_KEY") is not None
+
+
+# This is a hack to make sure that the openai api is set correctly
+# Right now it is overkill, but it will be useful when the api gets fixed and we no longer
+# hardcode the model names (i.e. OpenAI fixes their api)
```
pal03377, line 99: Is there some hope for the future to be able to remove this?

module_text_llm/module_text_llm/helpers/models/openai.py
```diff
@@ -0,0 +1,312 @@
...
+        else:
+            raise EnvironmentError(
+                "OpenAI api not available, please set LLM_OPENAI_API_KEY environment variable"
+            )
+
+    # API client is setup correctly
+    yield
+
+
+def _get_available_deployments(openai_models: dict[str, List[str]], model_aliases: dict[str, str]):
```
pal03377:
```suggestion
def _get_available_deployments(openai_models: Dict[str, List[str]], model_aliases: Dict[str, str]):
```
I usually prefer `dict`, but then you don't support Python 3.8. I'm not sure if we even want to, but only uppercasing all built-in types seems like a low price to pay for that, so I think we should keep them capitalized.

module_text_llm/module_text_llm/helpers/models/openai.py
```diff
@@ -0,0 +1,312 @@
...
+    available_deployments: dict[str, dict[str, Any]] = {
```
pal03377:
```suggestion
    available_deployments: Dict[str, Dict[str, Any]] = {
```

module_text_llm/module_text_llm/helpers/models/openai.py
```diff
@@ -0,0 +1,312 @@
...
+                for deployment in deployments:
+                    model_name = deployment.model
+                    if model_name in model_aliases:
+                        model_name = model_aliases[model_name]
+                    if model_name in openai_models["chat_completion"]:
+                        available_deployments["chat_completion"][deployment.id] = deployment
+                    elif model_name in openai_models["completion"]:
+                        available_deployments["completion"][deployment.id] = deployment
+                    elif model_name in openai_models["fine_tuneing"]:
+                        available_deployments["fine_tuneing"][deployment.id] = deployment
```
pal03377:
```suggestion
            deployments = openai.Deployment.list().get("data") or []  # type: ignore
            for deployment in deployments:
                model_name = deployment.model
                if model_name in model_aliases:
                    model_name = model_aliases[model_name]
                if model_name in openai_models["chat_completion"]:
                    available_deployments["chat_completion"][deployment.id] = deployment
                elif model_name in openai_models["completion"]:
                    available_deployments["completion"][deployment.id] = deployment
                elif model_name in openai_models["fine_tuneing"]:
                    available_deployments["fine_tuneing"][deployment.id] = deployment
```

Suggestion - I'm unsure if it's better, but it's one less indentation.

module_text_llm/module_text_llm/helpers/models/openai.py
```diff
@@ -0,0 +1,312 @@
...
+
+    return available_deployments
+
+
+def _get_available_models(openai_models, available_deployments):
+    available_models: dict[str, BaseLanguageModel] = {}
```
pal03377:
```suggestion
    available_models: Dict[str, BaseLanguageModel] = {}
```

module_text_llm/module_text_llm/helpers/models/openai.py
```diff
@@ -0,0 +1,312 @@
...
+
+    if openai_available:
+        openai_api_key = os.environ["LLM_OPENAI_API_KEY"]
+        for model_name in openai_models["chat_completion"]:
+            available_models[f"{OPENAI_PREFIX}{model_name}"] = ChatOpenAI(
```
pal03377: This looks simpler:
```suggestion
            available_models[OPENAI_PREFIX + model_name] = ChatOpenAI(
```

module_text_llm/module_text_llm/helpers/models/openai.py
```diff
@@ -0,0 +1,312 @@
...
+                model=model_name, openai_api_key=openai_api_key, client="")
+        for model_name in openai_models["completion"]:
+            available_models[f"{OPENAI_PREFIX}{model_name}"] = OpenAI(
```
pal03377:
```suggestion
            available_models[OPENAI_PREFIX + model_name] = OpenAI(
```

module_text_llm/module_text_llm/helpers/models/openai.py
```diff
@@ -0,0 +1,312 @@
...
+                model=model_name, openai_api_key=openai_api_key, client="")
+
+    if azure_openai_available:
+        azure_openai_api_key = os.environ["LLM_AZURE_OPENAI_API_KEY"]
+        azure_openai_api_base = os.environ["LLM_AZURE_OPENAI_API_BASE"]
+        azure_openai_api_version = os.environ["LLM_AZURE_OPENAI_API_VERSION"]
+
+        for deployment_name, deployment in available_deployments["chat_completion"].items():
+            available_models[f"{AZURE_OPENAI_PREFIX}{deployment_name}"] = AzureChatOpenAI(
```
pal03377:
```suggestion
            available_models[AZURE_OPENAI_PREFIX + deployment_name] = AzureChatOpenAI(
```

module_text_llm/module_text_llm/helpers/models/openai.py
```diff
@@ -0,0 +1,312 @@
...
+                deployment_name=deployment_name,
+                openai_api_base=azure_openai_api_base,
+                openai_api_version=azure_openai_api_version,
+                openai_api_key=azure_openai_api_key,
+                client="",
+            )
+
+        for deployment_name, deployment in available_deployments["completion"].items():
+            available_models[f"{AZURE_OPENAI_PREFIX}{deployment_name}"] = AzureOpenAI(
```
pal03377:
```suggestion
            available_models[AZURE_OPENAI_PREFIX + deployment_name] = AzureOpenAI(
```

module_text_llm/module_text_llm/helpers/models/openai.py
```diff
@@ -0,0 +1,312 @@
...
+        for deployment_name, deployment in available_deployments["chat_completion"].items():
+            available_models[f"{AZURE_OPENAI_PREFIX}{deployment_name}"] = AzureChatOpenAI(
+                deployment_name=deployment_name,
+                openai_api_base=azure_openai_api_base,
+                openai_api_version=azure_openai_api_version,
+                openai_api_key=azure_openai_api_key,
+                client="",
+            )
+
+        for deployment_name, deployment in available_deployments["completion"].items():
```
pal03377: Can't you loop over the key on `available_deployments` here to avoid writing the loop twice? `for deployment_type in "chat_completion", "completion":`

module_text_llm/module_text_llm/helpers/models/openai.py
```diff
@@ -0,0 +1,312 @@
...
+    ],
+    "fine_tuneing": [
+        "davinci",
+        "curie",
+        "babbage",
+        "ada",
+    ]
+}
+available_deployments = _get_available_deployments(
+    openai_models, _model_aliases)
```
pal03377:
```suggestion
available_deployments = _get_available_deployments(openai_models, _model_aliases)
```

module_text_llm/module_text_llm/helpers/models/openai.py
```diff
@@ -0,0 +1,312 @@
...
+available_models = _get_available_models(openai_models, available_deployments)
+
+
+OpenAIModel = Enum('OpenAIModel', {
+                   name: name for name in available_models})
```
pal03377:
```suggestion
OpenAIModel = Enum('OpenAIModel', {name: name for name in available_models})
```

module_text_llm/module_text_llm/helpers/models/openai.py
```diff
@@ -0,0 +1,312 @@
...
+
+
+if os.environ["LLM_DEFAULT_MODEL"] in available_models:
+    default_openai_model = OpenAIModel[os.environ["LLM_DEFAULT_MODEL"]]
+else:
+    default_openai_model = OpenAIModel["gpt-3.5-turbo"]
```
pal03377:
```suggestion
default_openai_model = OpenAIModel[os.environ.get("LLM_DEFAULT_MODEL", "gpt-3.5-turbo")]
```

module_text_llm/module_text_llm/helpers/models/openai.py
```diff
@@ -0,0 +1,312 @@
...
+
+
+# Long descriptions will be displayed in the playground UI and are copied from the OpenAI docs
+class OpenAIModelConfig(ModelConfig):
```
pal03377, line 234: Maybe add a docstring?

module_text_llm/module_text_llm/helpers/models/openai.py
```diff
@@ -0,0 +1,312 @@
...
+    def get_model(self) -> BaseLanguageModel:
+        model = available_models[self.model_name.value]
+        logger.debug("Using model %s", self.model_name.value)
+        for attr, value in self.dict().items():
+            model_kwargs = {}
+            if attr != "model_name":
+                if hasattr(model, attr):
+                    setattr(model, attr, value)
+                else:
+                    model_kwargs[attr] = value
```
pal03377: Can you add more comments about what you're doing here, please?

playground/src/components/base_info_header.tsx
```diff
@@ -11,6 +17,8 @@ type BaseInfoHeaderProps = {
   onChangeAthenaSecret: (value: string) => void;
   module: ModuleMeta | undefined;
   onChangeModule: (value: ModuleMeta) => void;
+  moduleConfig: any | undefined;
+  onChangeModuleConfig: (value: any | undefined) => void;
```
pal03377:
```suggestion
  onChangeModuleConfig: (value: any) => void;
```

playground/src/components/base_info_header.tsx
```diff
@@ -11,6 +17,8 @@ type BaseInfoHeaderProps = {
   onChangeAthenaSecret: (value: string) => void;
   module: ModuleMeta | undefined;
   onChangeModule: (value: ModuleMeta) => void;
+  moduleConfig: any | undefined;
```
pal03377:
```suggestion
  moduleConfig: any;
```
To my knowledge, `any` already allows for `undefined` values.

playground/src/components/form/description_field.tsx
```diff
@@ -0,0 +1,31 @@
...
+      </p>
+    );
+  } else {
+    return (
+      <div id={id} className='field-description'>
+        {description}
+      </div>
+    );
+  }
+}
```
pal03377, line 36: In case this is from somewhere else, please link to the source. If it is not: Wow! :D

playground/src/components/form/utils.ts
```diff
@@ -0,0 +1,28 @@
...
+  ValidatorType,
+  PathSchema,
+} from "@rjsf/utils";
+
+export function getUISchema(
+  validator: ValidatorType,
+  schema: RJSFSchema,
+  value: (property: string) => UIOptionsType
+) {
+  const paths = toPathSchema(validator, schema, undefined, schema);
```
pal03377, line 29: Please add a comment about what this does.

playground/src/components/module_config/default_schema_form.tsx
```diff
@@ -0,0 +1,46 @@
+import { getDefaultFormState } from "@rjsf/utils";
+import validator from "@rjsf/validator-ajv8";
+
+import Form from "@/components/form";
+import { ModuleConfigProps } from ".";
```
pal03377, line 5: What is this? I've never seen an import `from "."` in JS/TS 🤔

playground/src/components/module_config/module_llm.tsx
```diff
@@ -0,0 +1,59 @@
+import { getDefaultFormState } from "@rjsf/utils";
+import validator from "@rjsf/validator-ajv8";
+
+import Form from "@/components/form";
+import { getUISchema } from "@/components/form/utils";
+import { ModuleConfigProps } from ".";
+
+export default function ModuleLLMConfig({
```
pal03377, line 9: I don't immediately see which additional config UI elements this adds. Also, are you sure that it's not possible to render a good UI without any custom components? I really like the idea of "not caring" in the playground and just showing all available automatically gained options. In any case, an additional comment wouldn't hurt.

playground/src/components/module_config/index.tsx
```diff
@@ -0,0 +1,94 @@
...
+import ModuleLLMConfig from "./module_llm";
+import athenaFetcher from "@/helpers/athena_fetcher";
+
+// Add custom module config components here
+type CustomModuleConfig = "module_text_llm";
+const customModuleConfigComponents: {
+  [key in CustomModuleConfig]: React.FC<ModuleConfigProps>;
+} = {
+  module_text_llm: ModuleLLMConfig,
+};
```
pal03377, line 31: I'm a bit sad that this is needed 😞 
If it is really required, can you add some info on how to add something here? I suppose you have to rewrite `type CustomModuleConfig = "module_text_llm";` to `type CustomModuleConfig = "module_text_llm" | "my_new_module";`? If so, that's not so obvious.

playground/src/components/module_config/index.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+type SetConfig = Dispatch<SetStateAction<any | undefined>>;
+
+export type ModuleConfigProps = {
+  configOptions?: any;
+  moduleConfig: any | undefined;
```
pal03377:
```suggestion
  moduleConfig: any;
```

playground/src/components/module_requests/get_config_schema.tsx
```diff
@@ -0,0 +1,88 @@
...
+      })}`,
+      {
+        method: "GET",
+        headers: {
+          "Content-Type": "application/json",
+          "X-API-Secret": athenaSecret,
+          // No X-Module-Config here
+        },
+      }
+    );
```
pal03377, line 29: Not for this PR, but we should probably abstract away these kinds of requests in some later follow-up.

playground/src/components/module_requests/request_feedback_suggestions.tsx
```diff
@@ -42,6 +43,9 @@ async function requestFeedbackSuggestions(
         headers: {
           "Content-Type": "application/json",
           "X-API-Secret": athenaSecret,
+          ...(moduleConfig && {
+            "X-Module-Config": JSON.stringify(moduleConfig),
+          }),
```
pal03377, line 48: If `moduleConfig` is `undefined`, will the spread operator still work properly?

playground/src/components/module_requests/request_submission_selection.tsx
```diff
@@ -99,10 +104,10 @@ export default function SelectSubmission({
   } = useSubmissions(mode, exercise);
 
   const responseSubmissionView = (response: ModuleResponse | undefined) => {
-    if (!response || response.status !== 200 || typeof response.data !== "number") {
+    if (!response || response.status !== 200 || typeof response.data.data !== "number") {
```
pal03377:
```suggestion
    if (!response || response.status !== 200 || typeof response.data.data !== "number") {
```
It looks like a code smell to me that you must write `.data.data` here.

playground/src/helpers/athena_fetcher.ts
```diff
@@ -0,0 +1,45 @@
...
+    if (!res.ok) {
+      throw new FetchError(
+        "An error occurred while fetching the data.",
+        res.status,
+        await res.json()
+      );
+    }
+
+    return res.json();
+  };
```
pal03377: This looks like the reused fetching function I talked about earlier! Why is it not used everywhere?

playground/src/helpers/athena_fetcher.ts
```diff
@@ -0,0 +1,45 @@
...
+    super(message);
+    this.name = "FetchError";
+    this.status = status;
+    this.info = info;
+  }
+}
+
+const athenaFetcher =
+  (athenaSecret: string, moduleConfig: any | undefined = undefined) =>
+  async (url: string) => {
```
pal03377: Make this a class? Then you can also directly `export default class` it.

pal03377 (changes requested):
Wow, this is so incredibly cool! 🥳 I also really like most approaches you took in the code, keeping it extremely clean for module developers 💪 

It also worked (apart from an error for one text exercise, see comment below all the code review comments - https://github.com/ls1intum/Athena/pull/44#pullrequestreview-1527458787):

screenshot-2023-07-13_001565
screenshot-2023-07-13_001566
screenshot-2023-07-13_001567

I still have a large number of change requests, see below. Also, it would be good if future PRs could be smaller than this.

pal03377 (changes requested):
When I only send the submissions of test exercise 2 and then request feedback suggestions for submission 201 of that same exercise, I get this error: 
```
Traceback (most recent call last):
  File "/Users/paul/Sync/main/projects/Athena/module_text_llm/.venv/lib/python3.11/site-packages/anyio/streams/memory.py", line 98, in receive
    return self.receive_nowait()
           ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/paul/Sync/main/projects/Athena/module_text_llm/.venv/lib/python3.11/site-packages/anyio/streams/memory.py", line 93, in receive_nowait
    raise WouldBlock
anyio.WouldBlock

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/paul/Sync/main/projects/Athena/module_text_llm/.venv/lib/python3.11/site-packages/starlette/middleware/base.py", line 78, in call_next
    message = await recv_stream.receive()
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/paul/Sync/main/projects/Athena/module_text_llm/.venv/lib/python3.11/site-packages/anyio/streams/memory.py", line 118, in receive
    raise EndOfStream
anyio.EndOfStream

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/paul/Sync/main/projects/Athena/module_text_llm/.venv/lib/python3.11/site-packages/uvicorn/protocols/http/h11_impl.py", line 429, in run_asgi
    result = await app(  # type: ignore[func-returns-value]
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/paul/Sync/main/projects/Athena/module_text_llm/.venv/lib/python3.11/site-packages/uvicorn/middleware/proxy_headers.py", line 78, in __call__
    return await self.app(scope, receive, send)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/paul/Sync/main/projects/Athena/module_text_llm/.venv/lib/python3.11/site-packages/fastapi/applications.py", line 276, in __call__
    await super().__call__(scope, receive, send)
  File "/Users/paul/Sync/main/projects/Athena/module_text_llm/.venv/lib/python3.11/site-packages/starlette/applications.py", line 122, in __call__
    await self.middleware_stack(scope, receive, send)
  File "/Users/paul/Sync/main/projects/Athena/module_text_llm/.venv/lib/python3.11/site-packages/starlette/middleware/errors.py", line 184, in __call__
    raise exc
  File "/Users/paul/Sync/main/projects/Athena/module_text_llm/.venv/lib/python3.11/site-packages/starlette/middleware/errors.py", line 162, in __call__
    await self.app(scope, receive, _send)
  File "/Users/paul/Sync/main/projects/Athena/module_text_llm/.venv/lib/python3.11/site-packages/starlette/middleware/base.py", line 108, in __call__
    response = await self.dispatch_func(request, call_next)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/paul/Sync/main/projects/Athena/athena/athena/metadata.py", line 14, in dispatch
    response = await call_next(request)
               ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/paul/Sync/main/projects/Athena/module_text_llm/.venv/lib/python3.11/site-packages/starlette/middleware/base.py", line 84, in call_next
    raise app_exc
  File "/Users/paul/Sync/main/projects/Athena/module_text_llm/.venv/lib/python3.11/site-packages/starlette/middleware/base.py", line 70, in coro
    await self.app(scope, receive_or_disconnect, send_no_error)
  File "/Users/paul/Sync/main/projects/Athena/module_text_llm/.venv/lib/python3.11/site-packages/starlette/middleware/exceptions.py", line 79, in __call__
    raise exc
  File "/Users/paul/Sync/main/projects/Athena/module_text_llm/.venv/lib/python3.11/site-packages/starlette/middleware/exceptions.py", line 68, in __call__
    await self.app(scope, receive, sender)
  File "/Users/paul/Sync/main/projects/Athena/module_text_llm/.venv/lib/python3.11/site-packages/fastapi/middleware/asyncexitstack.py", line 21, in __call__
    raise e
  File "/Users/paul/Sync/main/projects/Athena/module_text_llm/.venv/lib/python3.11/site-packages/fastapi/middleware/asyncexitstack.py", line 18, in __call__
    await self.app(scope, receive, send)
  File "/Users/paul/Sync/main/projects/Athena/module_text_llm/.venv/lib/python3.11/site-packages/starlette/routing.py", line 718, in __call__
    await route.handle(scope, receive, send)
  File "/Users/paul/Sync/main/projects/Athena/module_text_llm/.venv/lib/python3.11/site-packages/starlette/routing.py", line 276, in handle
    await self.app(scope, receive, send)
  File "/Users/paul/Sync/main/projects/Athena/module_text_llm/.venv/lib/python3.11/site-packages/starlette/routing.py", line 66, in app
    response = await func(request)
               ^^^^^^^^^^^^^^^^^^^
  File "/Users/paul/Sync/main/projects/Athena/module_text_llm/.venv/lib/python3.11/site-packages/fastapi/routing.py", line 237, in app
    raw_response = await run_endpoint_function(
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/paul/Sync/main/projects/Athena/module_text_llm/.venv/lib/python3.11/site-packages/fastapi/routing.py", line 163, in run_endpoint_function
    return await dependant.call(**values)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/paul/Sync/main/projects/Athena/athena/athena/authenticate.py", line 38, in wrapper
    return await func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/paul/Sync/main/projects/Athena/athena/athena/endpoints.py", line 237, in wrapper
    feedbacks = await func(*args)
                ^^^^^^^^^^^^^^^^^
  File "/Users/paul/Sync/main/projects/Athena/module_text_llm/module_text_llm/__main__.py", line 37, in suggest_feedback
    return await suggest_feedback_basic(exercise, submission, config.approach, config.debug)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/paul/Sync/main/projects/Athena/module_text_llm/module_text_llm/suggest_feedback_basic.py", line 52, in suggest_feedback_basic
    credits = float(row["credits"])
              ^^^^^^^^^^^^^^^^^^^^^
TypeError: float() argument must be a string or a real number, not 'NoneType'
```

For another text exercise, this worked.

pal03377:
What I also currently don't like: The assessment module manager returns an object with `data` nested twice (because it wraps the module response, which wraps the return value of the function):
screenshot-2023-07-13_001568

playground/src/components/module_requests/request_feedback_suggestions.tsx
FelixTJDietrich: Yes, the undefined case was also the reason for doing it like this. 
image

athena/athena/endpoints.py
FelixTJDietrich: Done

env_example/module_text_llm.env
FelixTJDietrich: Will do :)

module_example/module_example/__main__.py
FelixTJDietrich: Hmm, I can do it for now. I will also add something that replaces `anyOf` with `oneOf` in the schema for union types. 

I thought maybe pydantic cannot cover all cases but for the near future this will probably be enough and also easier to understand. If someone runs into limitations they can replace it with a abstract class that supports `.schema()`

playground/src/components/module_requests/request_submission_selection.tsx
FelixTJDietrich: fixed

module_example/module_example/__main__.py
FelixTJDietrich: Nevermind about the `anyOf` for now, the `Union` type seems to work. It is much better now :)

module_text_llm/module_text_llm/config.py
FelixTJDietrich: Done

module_text_llm/module_text_llm/helpers/models/model_config.py
FelixTJDietrich: I want to do `.schema()`, it does support it but there are also things I don't want to configure and things that are omitted I think. This is just cleaner and I have more control of which configurations to expose.

Langchain is also a bit hacky and some config is not specified in the pydantic model (c.f. chat models `presence_penalty`, `top_p`, ...).

module_text_llm/module_text_llm/helpers/models/openai.py
FelixTJDietrich: Fixed it, now using `# pylint: disable=protected-access` (suddenly works for me)

module_text_llm/module_text_llm/helpers/models/openai.py
FelixTJDietrich: I will keep it for now like this.

What I might want to do in the future is to also allow dropping the `LLM_` prefix again, then you can either use azure or non-azure. But right now for development and evaluation having the option to use both is quite handy.

module_text_llm/module_text_llm/helpers/models/openai.py
FelixTJDietrich: Right now no, the openai azure and non-azure api are somewhat different but are exposed via one library `openai`. It would be amazing if I could import it twice as `openai` and `azure_openai`

FelixTJDietrich: I just saw that this is also possible with some trick to get around the caching. But then I would have the problem with langchain.

module_text_llm/module_text_llm/helpers/models/openai.py
FelixTJDietrich: Okay

module_text_llm/module_text_llm/helpers/models/openai.py
FelixTJDietrich: Good point

module_text_llm/module_text_llm/__main__.py
FelixTJDietrich: Done, it is now guaranteed :) Good idea

module_text_llm/module_text_llm/helpers/models/openai.py
FelixTJDietrich: I don't know why the formatter always does this...

module_text_llm/module_text_llm/helpers/models/openai.py
FelixTJDietrich: I changed it a bit, it should now be less buggy when using multiple configurations of the same model

playground/src/components/base_info_header.tsx
FelixTJDietrich: Yeah, I like the explicit undefined but I removed it everywhere now.

playground/src/components/form/description_field.tsx
FelixTJDietrich: Yeah it is slightly modified version of https://github.com/rjsf-team/react-jsonschema-form/blob/main/packages/core/src/components/templates/DescriptionField.tsx I will link it.

I wanted to enable markdown everywhere by default, so I added the `ReactMarkdown` around the string. They also have a way to enable markdown rendering, but it has some limitations... In case you enable markdown their way, it is using the `else` case ^^

playground/src/components/module_config/default_schema_form.tsx
FelixTJDietrich: This means that I'm importing from the same directory's `index.{js,jsx,ts,tsx}` file.

playground/src/components/module_config/module_llm.tsx
FelixTJDietrich: It is an opt-in way to customize your config form for the playground. There is a default form that is used and works out of the box, that renders without any custom components. For example `module_example` uses that form and any other future one.

By default, a field is rendered as input field which makes sense to me. For my llm modules, at least `module_text_llm` right now, I want to render a `textarea` for the prompt instead.

playground/src/components/module_config/index.tsx
FelixTJDietrich: I explained the reason for the customisation in the comment above. It is opt-in for when you are not satisfied with the default rendering of the json schema.

I added a comment on how to add your own customisation.

playground/src/helpers/athena_fetcher.ts
FelixTJDietrich: I thought about it but the problem is the `alert(...)` you are doing 😅
I can try to unify it

playground/src/helpers/athena_fetcher.ts
FelixTJDietrich: Hmm, I don't see the benefit of a class. I think the more functional approach is fine.

playground/src/helpers/athena_fetcher.ts
FelixTJDietrich: Nevermind, I remember it was actually the `POST` method that stopped me from it, I only needed a way to `GET` the schema. Let's do it in a folllow-up PR, I created an issue for it: #49

FelixTJDietrich:
The error you experienced with my module is still on my todo list. I will replace CSV parsing through JSON parsing and improve error handling so this does not happen anymore :S

assessment_module_manager/assessment_module_manager/module/request_to_module.py
```diff
@@ -11,13 +11,15 @@
 from assessment_module_manager import env
 
 T = TypeVar('T')
```
pal03377: Maybe rename `T` to `D` so that it's clearer that one is for Data and the other for Metadata?

assessment_module_manager/assessment_module_manager/module/request_to_module.py
```diff
@@ -32,23 +34,37 @@ async def find_module_by_name(
...
     """
     Helper function to send a request to a module.
     It raises appropriate FastAPI HTTPException if the request fails.
     """
     module_secret = env.MODULE_SECRETS[module.name]
+    headers = {'X-API-Secret': module_secret} if module_secret else None
+    if module_config:
+        if headers is None:
+            headers = {}
+        headers['X-Module-Config'] = module_config
```
pal03377:
```suggestion
headers = {}
if module_secret:
    headers['X-API-Secret'] = module_secret
if module_config:
    headers['X-Module-Config'] = module_config
```

Does something break if `headers` is `{}` instead of None? If no, the above simplification looks slightly better to me.

assessment_module_manager/assessment_module_manager/module/request_to_module.py
```diff
@@ -32,23 +34,37 @@ async def find_module_by_name(
...
+
     try:
         async with httpx.AsyncClient(base_url=module.url, timeout=600) as client:
-            response = await client.post(path, json=data, headers={'X-API-Secret': module_secret} if module_secret else None)
+            if method == "POST":
+                response = await client.post(path, json=data, headers=headers)
+            else:
+                response = await client.get(path, headers=headers)
```
pal03377:
```suggestion
            if method == "POST":
                response = await client.post(path, json=data, headers=headers)
            elif method == "GET":
                response = await client.get(path, headers=headers)
            else:
                raise ...
```
`PUT` and `DELETE` requests are also possible. These would currently just get converted into `GET` requests.

assessment_module_manager/assessment_module_manager/module/request_to_module.py
```diff
@@ -32,23 +34,37 @@ async def find_module_by_name(
...
     try:
         response.raise_for_status()
     except httpx.HTTPStatusError:
-        try:
-            response_data = response.json()
-        except json.JSONDecodeError:
-            response_data = response.text
-        return ModuleResponse(module_name=module.name, status=response.status_code, data=response_data)
-    return ModuleResponse(module_name=module.name, status=response.status_code, data=response.json())
+        pass
```
pal03377: I don't get this change. As far as I see, the new code segment here never does anything (either it raises and then catches or not).

assessment_module_manager/assessment_module_manager/module/request_to_module.py
```diff
@@ -32,23 +34,37 @@ async def find_module_by_name(
...
+
+    try:
+        response_data = response.json()
+        meta = response_data.get('meta', {})
+        response_data = response_data.get('data', response_data)
+    except json.JSONDecodeError:
+        response_data = response.text
```
pal03377, line 66: Should there be a log here?

athena/athena/endpoints.py
```diff
@@ -23,52 +29,128 @@
...
+    Callable[[E, List[S]], Coroutine[Any, Any, None]],
+    Callable[[E, List[S], C], Coroutine[Any, Any, None]]
+]):
     """
     Receive submissions from the Assessment Module Manager.
     The submissions consumer is usually called whenever the deadline for an exercise is reached.
+    
+    This decorator can be used with several types of functions: synchronous or asynchronous, with or without a module config.
+
+    Examples:
```
pal03377, line 44: Very nice! When I suggested these comments, I imagined them to be inline and next to the types, but this is even better 🎉

athena/athena/endpoints.py
```diff
@@ -96,64 +182,168 @@ async def wrapper(
...
+        return None
     return wrapper
 
 
-def feedback_provider(func: Union[Callable[[E, S], List[F]], Callable[[E, S], Coroutine[Any, Any, List[F]]]]):
+def feedback_provider(func: Union[
+    Callable[[E, S], List[F]],
+    Callable[[E, S, C], List[F]],
+    Callable[[E, S], Coroutine[Any, Any, List[F]]],
+    Callable[[E, S, C], Coroutine[Any, Any, List[F]]]
```
pal03377: I just noticed that the order here differs from the examples' order. Can you re-order the types?

athena/athena/endpoints.py
```diff
@@ -96,64 +182,168 @@ async def wrapper(
...
+        cls()
+    except ValidationError as exc:
+        raise TypeError(f'Cannot initialize {cls.__name__} without parameters, please provide default values for all parameters') from exc
+
+    @app.get("/config_schema")
+    async def wrapper():
+        return cls.schema()
+
+    app.state.config_schema_defined = True
+    return cls
```
pal03377, line 349: It's probably better if you don't change this, but to boast the knowledge I gained last semester in the `Advanced Python Programming` course a bit: This looks like a perfect use case for a metaclass :) You would define the metaclass and a base class to inherit from. I'll leave 100% of the choice whether you want to do this to you; metaclasses can be quite a complex beast, so maybe the code wouldn't even improve.

athena/athena/metadata.py
```diff
@@ -0,0 +1,109 @@
...
+    {
+        "data": {
+            "item_id": 42
+            "name": "Fancy Item"
+        },
+        "meta": {
+            "item_id": 42
+        }
+    }
+"""
```
pal03377, line 29: Wow, that's a really good docstring!

module_example/module_example/__main__.py
```diff
@@ -21,6 +28,13 @@ def receive_submissions(exercise: Exercise, submissions: List[Submission]):
     # Do something with the submissions
     logger.info("Doing stuff")
 
+    # Example use module config
+    # If you are not using module_config for your module, you can remove it from the function signature
+    logger.info("Config: %s", module_config)
+    if module_config.debug:
+        emit_meta('debug', True)
+        emit_meta('_comment', 'You can add any meta data you want here')
```
pal03377: Why is it `_comment` and not just `_comment` here?

module_example/module_example/__main__.py
```diff
@@ -48,17 +62,23 @@ def process_incoming_feedback(exercise: Exercise, submission: Submission, feedba
...
-def suggest_feedback(exercise: Exercise, submission: Submission) -> List[Feedback]:
+def suggest_feedback(exercise: Exercise, submission: Submission, module_config: Configuration) -> List[Feedback]:
     logger.info("suggest_feedback: Suggestions for submission %d of exercise %d were requested", submission.id, exercise.id)
     # Do something with the submission and return a list of feedback
+
+    # Example use of module config
+    # If you are not using module_config for your module, you can remove it from the function signature
+    logger.info("Config: %s", module_config)
+    if module_config.debug:
+        emit_meta('costs', '100.00€')
```
pal03377:
```suggestion
        emit_meta("costs", "100.00€")
```
consistency

module_text_llm/module_text_llm/helpers/models/openai.py
pal03377: :( ok

I'll close my eyes for a second 🙈

module_text_llm/module_text_llm/prompts/suggest_feedback_basic.py
```diff
@@ -2,7 +2,7 @@
 You are an AI tutor at a prestigious university tasked with grading and providing feedback to text assignments.
 
 VERY IMPORTANT: Effective feedback for text assignments should be:
-1. Constructive, 2. Specific, 3. Balanced, 4. Clear and Concise, 5. Actionable, 6. Educational, 7. Contextual
+1. Constructive, 2. Specific, 3. Balanced, 4. Clear and Concise, 5. Actionable, 6. Educational, 7. Contextual\
 """
```
pal03377, line 6: TIL that that's possible 😮

playground/src/components/module_config/module_llm.tsx
pal03377: Ok, makes sense 👍

playground/src/components/module_requests/request_feedback_suggestions.tsx
pal03377: Ah, interesting!

playground/src/components/module_requests/request_submission_selection.tsx
```diff
@@ -59,6 +63,7 @@ async function requestSubmissionSelection(
         module_name: "Unknown",
         status: response.status,
         data: await response.text(),
+        meta: {},
```
pal03377: We should be able to omit this, right? We might need to change the `ModuleResponse` type instead.

playground/src/components/module_response_view.tsx
```diff
@@ -12,14 +12,23 @@ export default function ModuleResponseView({
...
   try {
     formattedData = JSON.parse(formattedData);
   } catch (e) {
     // ignore
   }
+  try {
+    formattedMeta = JSON.parse(formattedMeta);
+  } catch (e) {
+    // ignore
+  }
```
pal03377:
```suggestion
  try {
    formattedData = JSON.parse(formattedData);
  } catch (e) {
    // continue to show the non-formatted response string
  }
  try {
    formattedMeta = JSON.parse(formattedMeta);
  } catch (e) {
    // continue to show the non-formatted response string
  }
```

Improving my own old comment here, I believe.

playground/src/model/module_response.ts
```diff
@@ -2,6 +2,7 @@ type ModuleResponse = {
   module_name: string;
   status: number;
   data: any;
+  meta: any;
```
pal03377:
```suggestion
  meta?: any;
```
Does this make all the `meta` changes unnecessary?

module_example/module_example/__main__.py
pal03377:
```suggestion
        emit_meta('_comment', 'You can add any metadata you want here')
```

pal03377 (changes requested):
Nice changes! I only have larger problems in `request_to_module.py` left now. I also tested it locally again.

Also, I notice that `endpoints.py` is getting more and more complex (which is not really your fault). We might need to think of some refactoring in the future.

assessment_module_manager/assessment_module_manager/module/request_to_module.py
FelixTJDietrich: That is probably fine, I will test it. Looks also better to me

athena/athena/endpoints.py
FelixTJDietrich: Reordered

assessment_module_manager/assessment_module_manager/module/request_to_module.py
FelixTJDietrich: I guess I will remove `raise_for_status` then?

athena/athena/endpoints.py
FelixTJDietrich: I looked it up and you are right. But this does basically the same and seems to be easier to understand. Let's keep it like this, also you don't have to use `metaclass=...` as a module developer that way.

assessment_module_manager/assessment_module_manager/module/request_to_module.py
FelixTJDietrich: I added one

module_example/module_example/__main__.py
FelixTJDietrich: I removed the underscore now

playground/src/components/module_requests/request_submission_selection.tsx
FelixTJDietrich: Done, thanks for the suggestion

assessment_module_manager/assessment_module_manager/module/request_to_module.py
pal03377: Yes 👍

pal03377 (approved):
Code looks good & I tested it again locally. Thanks for the contribution, very cool changes! I'll merge it right now

PR #46, opened 2023-07-04
Redirect root route to playground

pal03377:
### Motivation and Context

When directly visiting the Athena test server URL (e.g. by clicking the "View deployment" button in GitHub) or by just entering the URL normally, you get a 404 error. This is not so helpful. Therefore, we want to redirect to the only page with an actual UI: The playground.
screenshot-2023-07-04_001486

### Description

This PR adds a temporary redirect from `/` to `/playground`. This way, we can later change the root without problems if needed.

### Steps for Testing

- Deploy this PR
- Go to https://athenetest1-03.ase.cit.tum.de/
- Verify the redirect to the playground.

### DIFF
--- a/Caddyfile
+++ b/Caddyfile
@@ -2,6 +2,7 @@
 	email paul.schwind@tum.de
 }
 {$ATHENA_DOMAIN} {
+	redir / /playground 302
 	reverse_proxy /playground* http://playground:3000
 	reverse_proxy http://assessment_module_manager:5000
 }


pal03377:
I had weird problems when deploying this (redirect did not work sometimes), but now it should work: https://athenetest1-03.ase.cit.tum.de/

FelixTJDietrich (approved):
Great change :) Worked as expected for me

PR #47, opened 2023-07-06
Artemis integration

pal03377:
Related: https://github.com/ls1intum/Artemis/pull/6861

### Motivation and Context

Some changes are needed to integrate Athena into Artemis.

### Description

Change a few things to make the integration work (also some general improvements):
- Remove the `AvailableModuleNames` enum because it always causes issues when some modules are unavailable while the assessment module manager is starting. Replaced it with a simple string.
- Add an HTTP Auth protected `/logs` endpoint to see Docker logs => easier testing without SSH test server access
- Change `X-API-Secret` header to `Authorization` for consistency within Artemis
- Add a cleanup step to deployment so that the test server disk doesn't get overloaded with cached docker images
- Download CoFee config files on the test server on deployment
- Use PostgreSQL database on the test server instead of SQLite for a more production-like environment
- Let `/feedback` endpoint receive multiple feedback objects at once instead of one at a time
- Make a few fields nullable because they are nullable in Artemis
- Rename `TextSubmission.content` to `TextSubmission.text` for consistency
- Add `TextSubmission.language` field that is important for CoFee (which only works on English submissions)
- Accept any incoming objects with camelCase attribute names instead of snake_case as well => avoid snake_case names in Artemis DTOs
- Make the CoFee MongoDB docker service a bit quieter because you couldn't find anything in the logs

### Steps for Testing

Test https://github.com/ls1intum/Artemis/pull/6861 using the instructions there

### Screenshots

see https://github.com/ls1intum/Artemis/pull/6861


playground/src/model/exercise.ts
```diff
@@ -7,7 +7,7 @@ type ExerciseBase = {
   type: ExerciseType;
   max_points: number;
   bonus_points: number;
-  grading_instructions: string;
+  grading_instructions: string | null;
```
FelixTJDietrich: I like `undefined` more than `null`, idk what is really best practices but I always check for `undefined` instead of `null`. I guess semantically, `null` is more sensible but I still prefer `undefined`. 
```suggestion
  grading_instructions?: string;
```

playground/src/hooks/athena/send_feedback.ts
```diff
@@ -24,7 +24,7 @@ export function useSendFeedback(
   const athenaFetcher = useAthenaFetcher();
   return useMutation({
     mutationFn: async ({ exercise, submission, feedback }) => {
-      return await athenaFetcher("/feedback", { exercise, submission, feedback });
```
FelixTJDietrich, line 27: We can also update the module requests "Send Feedback". There we are doing a request for each feedback, but I can also do this in my human evaluation branch.

FelixTJDietrich (approved):
Changes make sense to me and work as expected on the Athena testserver.

playground/src/model/exercise.ts
pal03377: Makes sense! It's also the preferred choice in Artemis according to our guidelines, so we probably want to also do this 👍

playground/src/hooks/athena/send_feedback.ts
pal03377: Let's do that in a follow-up then 👍

PR #56, opened 2023-07-20
Improve feedback and some fixes

FelixTJDietrich:
### Motivation and Context

I'm having some issues with the feedback while implementing the human evaluation mode. This branch aims to fix some of those issues and improve on the feedback schema.

### Description

- [x] Change feedback schema `text`->`title` and `detail_text`->`description` (less confusing)
- [x] Fix potential feedback id collisions with the LMS ids
- [x] Return internal ids with the feedback suggestions 
- [x] Fix TypeScript feedback types
- [x] Improve TypeScript type import consistency
- [x] Adjust example data to use new feedback schema
- [x] Adjust export script to use the new feedback schema

### Open Issues
- [ ] Should I remove `title` from the feedback schema?

### Steps for Testing

1. Reset your db (@pal03377 are we planning some migration solution in the future?)
2. Successfully send feedback to Athena
3. Request feedback suggestions 
    - Should respond with the internal ids


FelixTJDietrich:
@pal03377 should I also drop the `feedback.title` from the schema? 
We are not really using it 🤔

playground/data/example/exercise-1/submissions/101/BinarySearch.java
```diff
@@ -13,7 +13,7 @@ public int performSearch(List<Chapter> chapters, String chapterName) {
             if (result == 0)
                 return chapters.get(theMiddle).getPageNumber();
             if (result > 0)
-                small = theMiddle + 1;
+                small = theMiddle + 1
```
FelixTJDietrich, line 16: Removed semicolon to match the feedback comment

athena/athena/models/db_feedback.py
```diff
@@ -1,12 +1,15 @@
...
 from .model import Model
 
 
 class DBFeedback(Model):
+    __table_args__ = (UniqueConstraint('lms_id'),)
+
     id = Column(Integer, primary_key=True, index=True, autoincrement=True)
-    detail_text = Column(String, nullable=False)
-    text = Column(String, nullable=False)
+    lms_id = Column(Integer, nullable=True)
```
pal03377, line 10: That's interesting! Does that mean that we will be able to use one Athena instance with multiple LMSes?

pal03377: Oh, now I see the `lms_id` unique constraint. Why do we need that? 🤔

pal03377 (approved):
Tested it locally and didn't have any issue. Also I looked at the code and everything I see looks like a total improvement 💪  Good job!

pal03377:
> @pal03377 should I also drop the `feedback.title` from the schema? We are not really using it 🤔

Let's discuss this separately from this PR

athena/athena/models/db_feedback.py
FelixTJDietrich: I guess you can do a combined uniqueness thingy if you want to allow multiple lms'. As of now I think we should just have one Athena instance per LMS

PR #59, opened 2023-07-24
Show 422 errors in the log

pal03377:
### Motivation and Context

When there are problems with the format in which Artemis sends data to Athena, it's hard to debug the problem.

### Description

Show details on 422 errors in the logs. This way, we can see what's wrong with the sent format.

### Steps for Testing

- Run the `assessment_module_manager` and the `module_example` locally.
- Run the following commands:
```
# Assessment Module Manager
curl http://localhost:5000/modules/programming/module_example/submissions --data-raw wrong
# Example Module
curl http://localhost:5001/submissions --data-raw wrong
```
- Check that both servers show the 422 problem in the logs.

### Screenshots

screenshot-2023-07-24_001678
screenshot-2023-07-24_001679

### DIFF
--- a/assessment_module_manager/assessment_module_manager/app.py
+++ b/assessment_module_manager/assessment_module_manager/app.py
@@ -1,4 +1,8 @@
-from fastapi import FastAPI
+from fastapi import FastAPI, Request
+from fastapi.exceptions import RequestValidationError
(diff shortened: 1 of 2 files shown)


FelixTJDietrich (approved):
Code LGTM and works as described 🎉

PR #61, opened 2023-07-24
More usable submission detail for evaluation mode

FelixTJDietrich:
### Motivation and Context

QoL improvements for viewing and assessing a submission and its feedback for both text and programming exercises. 
This is in preparation of the human evaluation mode and will be used to annotate the results.

### Description

Improves submission and exercise detail for text and programming exercises by integrating the Monaco text/code editor.
Displays the feedback inline and make it easy to view, add, or edit feedback.

❗️We have to remove Feedback Editor Test before merging, you can use it for testing the editor's edit feature:
image

For now the feedback looks like this:
image

**TODO in follow-up PR:**
 - Automatic evaluation results row below the feedback
    - E.g. rating of non-functional requirements (helpfulness, correctness, verboseness, ...)
    - Human annotator can then 👎👍 or rate how much they agree with the automatic evaluation

### Steps for Testing

> If you have previously exported some data from Artemis there are some issues with it, i.e. feedback references are off by one. I fixed it in #64, just in case you want to test with the evaluation data.

1. Use the Feedback Editor Test to see feedback editing in action
    1.1 Test text exercises
    1.2 Test programming exercises
2. Check submission details and exercise details if they look correct (you can't edit feedback there, inline feedback is shown in `send feedback` and `receive feedback suggestions`)

### Screenshots

*Programming Feedback (Editor)*
Screenshot 2023-07-30 at 11 09 24

*Text Feedback (Editor)*
image


playground/src/components/details/editor/code_editor.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+import { useFetchAndUnzip } from "@/helpers/fetch_and_unzip";
+
+import FileTree from "./file_tree";
+import FileEditor from "./file_editor";
+
+type CodeEditorProps = {
+  repository_url: string;
+  feedbacks?: Feedback[];
+  onFeedbacksChange?: (feedback: Feedback[]) => void;
```
pal03377: I feel like it's a bit weird that the `repository_url` is `snake_case`-d and `onFeedbacksChange` is not. Even though our data on the Python side takes it in this format, maybe we should have it as `repositoryUrl` here regardless

playground/src/components/details/editor/code_editor.tsx
```diff
@@ -0,0 +1,74 @@
...
+  createNewFeedback?: () => Feedback;
+};
+
+export default function CodeEditor({
+  repository_url,
+  feedbacks,
+  onFeedbacksChange,
```
pal03377: same here with `repository_url` ^^

playground/src/components/details/editor/editor_widget.tsx
```diff
@@ -0,0 +1,97 @@
...
+export function EditorWidget({ editor, children, afterLineNumber, afterColumn, modelPath }: EditorWidgetProps) {
+  const id = useId();
+  const portalNode = useMemo(() => createHtmlPortalNode(), []);
+  const resizeObserverRef = useRef<ResizeObserver>();
+  const overlayWidgetRef = useRef<editor.IOverlayWidget>();
+  const zoneIdRef = useRef<string>();
+  
+  const updateWidget = () => {
+    overlayWidgetRef.current && editor.removeOverlayWidget(overlayWidgetRef.current);
+    zoneIdRef.current && editor.changeViewZones((accessor) => accessor.removeZone(zoneIdRef.current!));
```
pal03377:
```suggestion
    if (overlayWidgetRef.current) {
        editor.removeOverlayWidget(overlayWidgetRef.current);
    }
    if (zoneIdRef.current) {
        editor.changeViewZones((accessor) => accessor.removeZone(zoneIdRef.current));
    }
```
(I don't know if omitting the `!` in `zoneIdRef.current!` really works then in TS)

playground/src/components/details/editor/editor_widget.tsx
```diff
@@ -0,0 +1,97 @@
...
+ 
+  useEffect(() => {
+    updateWidget();
+
+    // Last resort workaround for the widget not being created
+    const observer = new MutationObserver((mutationsList, observer) => {
+      // If the widget doesn't exist, update it
+      if (!document.getElementById(`widget-${id}-zone`)) {
+        updateWidget();
+      }
```
pal03377: This doesn't really make sense to me logically. How can you update something if it does not exist? Should it be `createWidget` maybe? 🤔

playground/src/components/details/editor/file_editor.tsx
```diff
@@ -0,0 +1,370 @@
...
+  const feedbacksToRender = feedbacks?.filter((feedback) => {
+    const referenceType = getFeedbackReferenceType(feedback);
+    if (referenceType === "unreferenced") {
+      return false;
+    }
+    if ("file_path" in feedback) {
+      return feedback.file_path === filePath;
+    } else {
+      return true;
+    }
```
pal03377:
```suggestion
    }
    return true;
```

playground/src/components/details/editor/file_editor.tsx
```diff
@@ -0,0 +1,370 @@
...
+          });
+          const endIndex = model.getOffsetAt({
+            lineNumber: hoverPosition.lineNumber,
+            column: model.getLineMaxColumn(hoverPosition.lineNumber),
+          });
+          newFeedback = {
+            ...newFeedback,
+            index_start: startIndex,
+            index_end: endIndex,
+          };
```
pal03377:
```suggestion
          newFeedback.index_start = model.getOffsetAt({
            lineNumber: hoverPosition.lineNumber,
            column: model.getLineMinColumn(hoverPosition.lineNumber),
          });
          newFeedback.index_end = model.getOffsetAt({
            lineNumber: hoverPosition.lineNumber,
            column: model.getLineMaxColumn(hoverPosition.lineNumber),
          });
```
I think this approach would also be more readable in the other places here

playground/src/components/details/editor/code_editor.tsx
pal03377: Ah, I see: The problem is that our data models in TS also have underscores. Maybe we should do a PR in the future where we generally change this, considering that Athena now also accepts camelCase input values

playground/src/components/details/submission_detail/index.tsx
```diff
@@ -1,29 +1,52 @@
...
+      title: "",
+      description: "",
+      type: submission.type,
+      exercise_id: submission.exercise_id,
+      submission_id: submission.id,
+      is_new: true,
+      meta: {},
+    };
+    return newFeedback;
+  };
```
pal03377: I would move this into its own file / some utils file and only pass the `submission` to the components that need to create feedback

playground/src/components/details/submission_detail/programming.tsx
```diff
@@ -1,26 +1,55 @@
...
+        <div className="space-y-2 mt-5">
+          <h3 className="ml-2 text-lg font-medium">Unreferenced Feedback</h3>
+          {feedbacks?.map((feedback) => (
+            getFeedbackReferenceType(feedback) === "unreferenced" && (
+            <InlineFeedback
+              key={feedback.id}
+              feedback={feedback}
+              onFeedbackChange={
+                onFeedbacksChange &&
+                getOnFeedbackChange(feedback, feedbacks, onFeedbacksChange)
```
pal03377, line 39: I feel like you could just update `feedbacks` in your state (which you probably have to do in some form anyways) and then call `onFeedbacksChange(feedbacks)` directly. Then, there wouldn't be the need for this somewhat weird `getOnFeedbackChange` transformation function 🤔

playground/src/components/module_requests/feedback_editor_test.tsx
```diff
@@ -0,0 +1,92 @@
...
+    if (submission && loadedFeedbacks) {
+      setFeedbacks(
+        loadedFeedbacks.filter((f) => f.submission_id === submission.id)
+      );
+    }
+  }, [submission, loadedFeedbacks]);
+
+  useEffect(() => {
+    setSubmission(undefined);
+  }, [exercise]);
```
pal03377: This feels a bit dangerous. Can you somehow more explicitly remove the submission only when it is really needed?

playground/src/components/module_requests/feedback_editor_test.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+  useEffect(() => {
+    setExercise(undefined);
+  }, [module, mode]);
```
pal03377: Same here, it seems a bit error prone to always reset the exercise when module or mode change. Maybe you can hook into an onChange below?

playground/src/components/module_requests/index.tsx
```diff
@@ -6,10 +6,14 @@ import SendFeedback from "@/components/module_requests/send_feedback";
 import RequestFeedbackSuggestions from "@/components/module_requests/request_feedback_suggestions";
 import SelectSubmission from "@/components/module_requests/request_submission_selection";
 
+// TODO: REMOVE THIS BEFORE MERGING TO MAIN
+import FeedbackEditorTest from "./feedback_editor_test";
```
pal03377, line 10: Review comment for todo

playground/src/components/module_requests/index.tsx
```diff
@@ -6,10 +6,14 @@ import SendFeedback from "@/components/module_requests/send_feedback";
...
+
 export default function ModuleRequests({ module }: { module: ModuleMeta }) {
   return (
     <>
       <h2 className="text-4xl font-bold text-white mb-4">Module Requests</h2>
+      <FeedbackEditorTest module={module} /> {/* TODO: REMOVE THIS BEFORE MERGING TO MAIN */}
```
pal03377, line 16: Review comment for todo

playground/src/helpers/get_data.ts
```diff
@@ -72,6 +72,65 @@ function replaceJsonPlaceholders(
...
+  const submissions = json.submissions?.map((submissionJson: any) => {
+    submissionJson.type = exerciseType;
+    const feedbacks = submissionJson.feedbacks?.map((feedbackJson: any) => {
+      feedbackJson.type = exerciseType;
+      return feedbackJson;
+    });
+    submissionJson.feedbacks = feedbacks;
+    return submissionJson;
+  });
+  json.submissions = submissions;
```
pal03377:
```suggestion
  json.submissions = json.submissions?.map((submissionJson: any) => {
    submissionJson.type = exerciseType;
    submissionJson.feedbacks = submissionJson.feedbacks?.map((feedbackJson: any) => {
      feedbackJson.type = exerciseType;
      return feedbackJson;
    });
    return submissionJson;
  });
```

playground/src/model/feedback.ts
```diff
@@ -1,53 +1,127 @@
...
+      return undefined;
+    }
+    const startPosition = model.getPositionAt(feedback.index_start ?? feedback.index_end!);
+    const endPosition = model.getPositionAt(feedback.index_end ?? feedback.index_start!);
+    return {
+      startLineNumber: startPosition.lineNumber,
+      startColumn: startPosition.column,
+      endLineNumber: endPosition.lineNumber,
+      endColumn: endPosition.column,
+    }
```
pal03377:
```suggestion
    };
```

playground/src/model/feedback.ts
```diff
@@ -1,53 +1,127 @@
...
+  if (feedback.type === "programming") {
+    if (feedback.file_path !== undefined && feedback.line_start !== undefined) {
+      return "referenced";
+    } else if (feedback.file_path !== undefined) {
+      return "unreferenced_file";
+    }
+  } else if (feedback.type === "text") {
+    if (feedback.index_start !== undefined && feedback.index_end !== undefined) {
+      return "referenced";
+    }
```
pal03377: So a text feedback with only the start index is also unreferenced? Is this correct? 🤔

playground/src/model/feedback.ts
```diff
@@ -1,53 +1,127 @@
...
+export function getOnFeedbackChange(feedback: Feedback, feedbacks: Feedback[], onFeedbacksChange: (feedbacks: Feedback[]) => void): (newFeedback: Feedback | undefined) => void {
+  return (newFeedback: Feedback | undefined) => {
+    let newFeedbacks = [...feedbacks];
+    if (newFeedback === undefined) {
+      newFeedbacks = newFeedbacks.filter((f) => f.id !== feedback.id);
+    } else {
+      newFeedbacks = newFeedbacks.map((f) => f.id === feedback.id ? newFeedback : f);
+    }
+    onFeedbacksChange(newFeedbacks);
+  }
```
pal03377:
```suggestion
  };
```

pal03377 (commented):
Code looks mostly good, I have a few suggestions though

playground/src/model/feedback.ts
FelixTJDietrich: I assumed that we always have an `index_start` and `index_end`, does it make sense to have zero width feedback? For programming langauges it is different, because there we give feedback on at least a whole line.

playground/src/components/details/editor/code_editor.tsx
FelixTJDietrich: I changed it nor to `repositoryUrl` and created issue #68 for a potential follow-up.

playground/src/components/details/editor/editor_widget.tsx
FelixTJDietrich: Changed. JS supports ! as far as I know, for example you can so `var?.someFunc()`, so it also makes sense to have `var!.someFunc()` which should throw an error AFAIK

playground/src/components/details/editor/editor_widget.tsx
FelixTJDietrich: I changed the comment. The widget is two parts, the view zone in the editor making space. This is in the background and non-interactable -> I need an overlay at the same location as the view zone. 

It is a bit hard to work with the lifecycle of React and Monaco, the view zones get destroyed when the model behind the text editor changes. There seems to be no way to hook into that lifecycle sadly :/ 

Anyways, the dirty fix is to just check if the view zone exists and update everything if it doesn't.

playground/src/components/details/editor/file_editor.tsx
FelixTJDietrich: I should do that more often 🤔 thanks!

playground/src/components/details/submission_detail/index.tsx
FelixTJDietrich: You are right, previously there was the dependency of the exerciseType but now I also have it in the submission so. I refactored it :)

playground/src/components/details/submission_detail/programming.tsx
FelixTJDietrich: Yeah I could just inline the function like this:
```ts
onFeedbacksChange &&
((newFeedback: Feedback | undefined) => {
  let newFeedbacks = [...feedbacks];
  if (newFeedback === undefined) {
    newFeedbacks = newFeedbacks.filter(
      (f) => f.id !== feedback.id
    );
  } else {
    newFeedbacks = newFeedbacks.map((f) =>
      f.id === feedback.id ? newFeedback : f
    );
  }
  onFeedbacksChange(newFeedbacks);
})
```

But this would just lead to a code duplication at three places. I guess you could argue that the name of the function is slightly weird. I will keep it like this for now.

playground/src/components/module_requests/feedback_editor_test.tsx
FelixTJDietrich: I moved this to the onChange of the exercise select 👍

playground/src/components/module_requests/feedback_editor_test.tsx
FelixTJDietrich: Hmm, there is no onModule or onMode change below. But I think you are right I can do it a bit safer. I will just reset the exercise and submission if the module type changes 👍

playground/src/components/details/editor/editor_widget.tsx
pal03377: That won't throw an error; it will only tell TS to shut up about type errors ;) https://stackoverflow.com/a/42274019/4306257 

Therefore, it has no effect here.

playground/src/model/feedback.ts
pal03377: Hmm, even if we have zero-width feedback, we should always have both indexes (in that case both set to the same value), or none. I think this makes sense 👍

playground/src/components/details/editor/editor_widget.tsx
pal03377: However, we get a type error here otherwise -> we need the `!`

pal03377:
screenshot-2023-08-11_001767

I get a 404 error on the test server 🤔  Do you know where this could come from?

FelixTJDietrich:
> screenshot-2023-08-11_001767
> I get a 404 error on the test server 🤔 Do you know where this could come from?

`req.headers["x-forwarded-proto"]` seems to be `https,http` for the deployed playground.

I'm trying to fix it with:
```ts
import type { NextApiRequest } from "next/types";

export default function getOriginFromRequest(req: NextApiRequest): string {
  const host = req.headers.host;
  let protocol: string | undefined;  
  const forwardedProto = req.headers["x-forwarded-proto"];
  if (typeof forwardedProto === 'string') {
    protocol = forwardedProto.split(',')[0];
  } else if (Array.isArray(forwardedProto)) {
    protocol = forwardedProto[0];
  } else {
    protocol = 'http';
  }
  return `${protocol}://${host}`;
}
```

FelixTJDietrich:
Fixed:
image

pal03377 (approved):
Looked at the changes since last time & tested on the test server 👍

PR #63, opened 2023-07-28
Only build necessary docker images

pal03377:
### Motivation and Context

Waiting for all Docker images to build before being able to deploy can get annoying, especially if the docker image for something you didn't even touch in your PR takes a long time.

### Description

Solution: We only re-build docker images when necessary. Building and uploading an image is required is required if one of the following conditions is true:
1. The Dockerfile in the directory has changed since the branch was created
2. The Docker image does not exist on GitHub Packages yet
3. The image is for a module, and the `athena` package was updated
4. The current branch is develop

### Steps for Testing

Add commits to this PR and observe which images are built (no need to reset after each commit, it should always take the difference to the last push):
- Add a commit with no changes
- Add a commit where one module or the playground is changed
- Add a commit where multiple modules are changed
- Add a commit where the `athena` package is changed

### Screenshots

No changes (compared to develop):
screenshot-2023-07-29_001741

### DIFF
--- a/.github/scripts/images-to-build.sh
+++ b/.github/scripts/images-to-build.sh
@@ -24,6 +24,11 @@ for DIR in */; do
     if [[ -e "${DIR}Dockerfile" ]]; then
         DIR=${DIR%/} # Remove trailing slash
 
+        # athena is always built and does not need to be re-built
+        if [[ "$DIR" == "athena" ]]; then
+            continue
+        fi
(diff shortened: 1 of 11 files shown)


FelixTJDietrich (approved):
Really nice! Code looks good to me and I followed your testing steps. All the changed containers were built 🎉

PR #69, opened 2023-08-10
Make feedback description optional

pal03377:
### Motivation and Context

When we tested https://github.com/ls1intum/Artemis/pull/6861 in a testing session, we found an issue where you could get an HTTP error 422 from Athena. According to @Strohgelaender, you can provoke this only as an instructor because then, some checks for feedback are disabled. We should prevent these kinds of issues, even if they only happen for instructors.

### Description

Make the `description` field on feedback optional.

### Steps for Testing

- Test feedback sending using the playground
- Maybe test sending feedback with no description


Strohgelaender (approved):

FelixTJDietrich (approved):
Code LGTM

PR #73, opened 2023-08-12
Add feedback editor to send feedbacks

FelixTJDietrich:
### Motivation and Context

#47 added some changes that allows to send feedbacks for a whole submission at a time, we did not update the playground to consider this.
#61 introduced a feedback editor for a submission which is perfectly suited for the send feedbacks under module requests, also the feedback editor debug view accidentally got merged into main. 

### Description

- Remove feedback editor test debug view
- Remove feedback selector and send feedbacks as batch per submission
- Enable feedback editor for the send feedbacks section

### Steps for Testing

1. Open the playground and navigate to Send Feedbacks
2. Play around with it to check if anything is not working as intended

### Screenshots

image


playground/src/components/module_requests/send_feedbacks.tsx
```diff
@@ -0,0 +1,254 @@
...
+  const [isAllSubmissions, setIsAllSubmissions] = useState<boolean>(true);
+  const [selectedSubmission, setSelectedSubmission] = useState<Submission | undefined>(undefined);
+
+  const [feedbacks, setFeedbacks] = useState<Feedback[]>([]);
+  const [removedFeedbackIds, setRemovedFeedbackIds] = useState<number[]>([]);
+  const [responses, setResponses] = useState<(ModuleResponse | undefined)[]>([]);
+
+  const setFeedbacksAndTrackChanges = (newFeedbacks: Feedback[]) => {
+    const storedFeedbackIds = new Set<number>(
+      storedFeedbacks?.map((f) => f.id) ?? []
```
pal03377, line 52: Where does `storedFeedbacks` come from? I don't really see how it can be defined at all here... 🤔

playground/src/components/module_requests/send_feedbacks.tsx
```diff
@@ -0,0 +1,254 @@
...
+  const { isLoading, error, mutateAsync, reset } = useSendFeedbacks();
+
+  const feedbacksAreChanged =
+    removedFeedbackIds.length > 0 || feedbacks.some((f) => f.isChanged);
+  const resetFeedbacks = () => {
+    setFeedbacks(storedFeedbacks ?? []);
+    setRemovedFeedbackIds([]);
+  };
+
+  // Handle resets with useEffect to avoid stale state
```
pal03377: What does this mean?

pal03377 (approved):
Thanks for the improvements & the fix after #47! 😅  Everything looks good to me and it works locally. There is an issue with sending feedback on the test server because of too large IDs, but this is not because of this PR. I'll solve it in a separate one.

playground/src/components/module_requests/send_feedbacks.tsx
FelixTJDietrich: `storedFeedbacks` comes from below maybe `fetchedFeedback` is a better name for it. Idk I think the stored naming also makes sense

playground/src/components/module_requests/send_feedbacks.tsx
FelixTJDietrich: I removed the comment now, it belonged to a different useEffect that I deleted

playground/src/components/module_requests/send_feedbacks.tsx
FelixTJDietrich: I moved the storedFeedback up now and added a comment.

FelixTJDietrich:
Still works on the test server, will merge it now.

PR #74, opened 2023-08-13
Find specific Docker tag for build

pal03377:
### Motivation and Context

#73 has issues where the necessary docker images are not built, even though they don't exist yet.

### Description

Solution: Properly detect if the images already exist or not. Before, the tag wasn't taken into consideration by accident. Now, the specific tag of the Docker image is checked in GitHub.

### Steps for Testing

Note that the build step to find images to build correctly identifies the image as missing: https://github.com/ls1intum/Athena/actions/runs/5849504932/job/15857908702?pr=74#step:3:45

Also, note that merging the branch of this PR into #73 fixes the issue there: All missing images are now built: https://github.com/ls1intum/Athena/actions/runs/5849540999/job/15857980555
(you can also check that the deployment now works correctly)

### Screenshots

screenshot-2023-08-13_001780

### DIFF
--- a/.github/scripts/images-to-build.sh
+++ b/.github/scripts/images-to-build.sh
@@ -57,20 +57,24 @@ for DIR in */; do
         fi
 
         # Construct Docker image name and tag
-        IMAGE_NAME="athena%2F$DIR"
+        IMAGE_NAME="athena/$DIR"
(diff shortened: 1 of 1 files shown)


FelixTJDietrich (approved):
Thanks for that, it seems to work on #73 :)

PR #75, opened 2023-08-14
Allow large IDs in the database

pal03377:
### Motivation and Context

When the IDs of exercises, submissions, or feedback get too large, we get a server error when sending the data to the server.

### Description

I increased the data size for all IDs to `BigInteger` instead of `Integer`.

### Steps for Testing
Send an exercise/submission/feedback to Athena with a large ID. This is easier to do after #73 is merged because it uses timestamps for the IDs of feedback.


FelixTJDietrich (approved):
Tested on the test server and it works as expected. Code also LGTM 
I will have to fix a bug with the send feedbacks section though :|

PR #78, opened 2023-08-17
Add LLaMA 2 through Replicate and improve config for text module

FelixTJDietrich:
### Motivation and Context

We want to evaluate on LLaMA 2 models and there were some errors with the environment when setting up the LLM api keys.

### Description

- Fix some issues with the environment in the OpenAI config
- Add Replicate as provider with the three LLaMA 2 chat models I want to evaluate on

### Issues
- [ ] When switching between providers, like OpenAI -> Replicate -> OpenAI some default parameters are unset
     - I don't know how to fix it, maybe just skip it. It is just a bit of a worse UX
- [x] Prospector union type error

### Steps for Testing

1. Setup some placeholder api key i.e. "XXXXXX" for replicate (or create an account and link credit card on https://replicate.com)
2. Check out the config schema in the playground
3. Optional: Module requests generate feedback suggestions with Replicate

### Screenshots

image


module_text_llm/module_text_llm/helpers/models/__init__.py
```diff
@@ -0,0 +1,36 @@
...
+    raise EnvironmentError(
+        "No model configurations available, please set up at least one provider in the environment variables.")
+
+if 'DefaultModelConfig' not in globals():
+    DefaultModelConfig = types[0]
+
+if len(types) == 1:
+    ModelConfigType = types[0]
+else:
+    ModelConfigType = Union[tuple(types)]  # type: ignore
```
FelixTJDietrich: @pal03377 do you have an idea how to fix this? :S

pal03377: I think the solution here is to not use types this dynamically and instead allow both values at all times. You will still have good autocompletion & checks, and you only "lose" the information that only one value is available each.

```suggestion
import os
from typing import Type, Union, List
from module_text_llm.helpers.models.model_config import ModelConfig
import module_text_llm.helpers.models.openai as openai_config
import module_text_llm.helpers.models.replicate as replicate_config


DefaultModelConfig: Type[ModelConfig]
default_model_name = os.environ.get("LLM_DEFAULT_MODEL")

types: List[Type[ModelConfig]] = []
try:
    types.append(openai_config.OpenAIModelConfig)
    if default_model_name in openai_config.available_models:
        DefaultModelConfig = openai_config.OpenAIModelConfig
except AttributeError:
    pass

try:
    types.append(replicate_config.ReplicateModelConfig)
    if default_model_name in replicate_config.available_models:
        DefaultModelConfig = replicate_config.ReplicateModelConfig
except AttributeError:
    pass

if not types:
    raise EnvironmentError(
        "No model configurations available, please set up at least one provider in the environment variables.")

if 'DefaultModelConfig' not in globals():
    DefaultModelConfig = types[0]

ModelConfigType = Union[openai_config.OpenAIModelConfig, replicate_config.ReplicateModelConfig]
```

module_text_llm/module_text_llm/helpers/models/openai.py
```diff
@@ -90,8 +90,8 @@ def _use_openai_credentials():
     openai.api_version = None
 
 
-openai_available = os.environ.get("LLM_OPENAI_API_KEY") is not None
-azure_openai_available = os.environ.get("LLM_AZURE_OPENAI_API_KEY") is not None
+openai_available = len(os.environ.get("LLM_OPENAI_API_KEY") or "") > 0
+azure_openai_available = len(os.environ.get("LLM_AZURE_OPENAI_API_KEY") or "") > 0
```
pal03377: I really like how explicit this is, so this is a purely optional suggestion:
```suggestion
openai_available = bool(os.environ.get("LLM_OPENAI_API_KEY"))
azure_openai_available = bool(os.environ.get("LLM_AZURE_OPENAI_API_KEY"))
```

module_text_llm/module_text_llm/helpers/models/replicate.py
```diff
@@ -0,0 +1,110 @@
...
+    # LLAMA 2 70B Chat
+    # https://replicate.com/replicate/llama-2-70b-chat
+    "llama-2-70b-chat": "replicate/llama-2-70b-chat:58d078176e02c219e11eb4da5a02a7830a283b14cf8f94537af893ccff5ee781",
+    # LLaMA 2 13B Chat
+    # https://replicate.com/a16z-infra/llama-2-13b-chat
+    "llama-2-13b-chat": "a16z-infra/llama-2-13b-chat:2a7f981751ec7fdf87b5b91ad4db53683a98082e9ff7bfd12c8cd5ea85980a52",
+    # LLaMA 2 7B Chat
+    # https://replicate.com/a16z-infra/llama-2-7b-chat
+    "llama-2-7b-chat": "a16z-infra/llama-2-7b-chat:7b0bfc9aff140d5b75bacbed23e91fd3c34b01a1e958d32132de6e0a19796e2c",
+}
```
pal03377, line 28: Can you add some documentation on how to update this if necessary?

module_text_llm/module_text_llm/helpers/models/replicate.py
```diff
@@ -0,0 +1,110 @@
...
+
+available_models = {}
+if len(os.environ.get("REPLICATE_API_TOKEN") or "") > 0:
```
pal03377:
```suggestion
if os.environ.get("REPLICATE_API_TOKEN"):  # If Replicate is available
```

pal03377 (commented):
I only looked at the code for now. It looks almost 100% good!

module_text_llm/module_text_llm/helpers/models/__init__.py
FelixTJDietrich: Hmm :/ But this will sadly mess up the JSON schema then, so you would be able to select replicate on the playground even though it is not configured for example.

FelixTJDietrich: Fixed it in a different way for now. Thanks!

module_text_llm/module_text_llm/helpers/models/replicate.py
FelixTJDietrich: Added some comment 👍

module_text_llm/module_text_llm/helpers/models/__init__.py
FelixTJDietrich: Fixed

FelixTJDietrich:
Somehow prospector did not like it without the ignore...

pal03377 (approved):

PR #79, opened 2023-08-18
Use OpenAI function calling and general improvements for `module_text_llm`

FelixTJDietrich:
This branch builds on #78 (mainly for testing reasons). Do not merge before that one is merged.

### Motivation and Context

Issues with the feedback generation and new features released by OpenAI and LangChain.

### Description

- Use OpenAI's function calling for getting structured output (they finetuned on json data for better performance 🎉)
    - Note: This only works for OpenAI's chat models (and also for the newer iterations, but we don't use the old ones)
- Include example solution
- Handle long prompts more gracefully (first omit example_solution, then problem_statement, and then fail)
- Use auto fixing output parsers for models that don't use function calling

### Steps for Testing

1. `cd module_text_llm`
2. `poetry install` (added `tiktoken` and updated `langchain`)
3. Run module and test the module's request feedback suggestions endpoint via the playground


module_text_llm/module_text_llm/generate_suggestions.py
```diff
@@ -0,0 +1,104 @@
...
+    num_tokens_from_prompt,
+    predict_and_parse
+)
+from module_text_llm.helpers.utils import add_sentence_numbers, get_index_range_from_line_range
+
+class FeedbackModel(BaseModel):
+    title: str = Field(..., description="Very short title, i.e. feedback category", example="Logic Error")
+    description: str = Field(..., description="Feedback description")
+    line_start: Optional[int] = Field(..., description="Referenced line number start, or empty if unreferenced")
+    line_end: Optional[int] = Field(..., description="Referenced line number end, or empty if unreferenced")
```
pal03377: Do you need the `...`?

module_text_llm/module_text_llm/helpers/llm_utils.py
```diff
@@ -0,0 +1,144 @@
...
+        max_input_tokens (int): Maximum number of tokens allowed
+        omittable_features (List[str]): List of features that can be omitted, ordered by priority (least important first)
+        debug (bool): Debug flag
+
+    Returns:
+        (dict, bool): Tuple of (prompt_input, should_run) where prompt_input is the input with omitted features and 
+                      should_run is True if the model should run, False otherwise
+    """
+    if num_tokens_from_prompt(prompt, prompt_input) <= max_input_tokens:
+        return prompt_input, True
```
pal03377, line 56:
```suggestion
        # Full prompt fits into LLM context => should run with full prompt
        return prompt_input, True
```

module_text_llm/module_text_llm/helpers/llm_utils.py
```diff
@@ -0,0 +1,144 @@
...
+def supports_function_calling(model: BaseLanguageModel):
+    """Returns True if the model supports function calling, False otherwise
+
+    Args:
+        model (BaseLanguageModel): The model to check
+
+    Returns:
+        boolean: True if the model supports function calling, False otherwise
+    """
+    return isinstance(model, ChatOpenAI)
```
pal03377, line 83: Why is this not solved with inheritance? I.e., an attribute on the `BaseLanguageModel` itself

module_text_llm/module_text_llm/helpers/llm_utils.py
```diff
@@ -0,0 +1,144 @@
...
+    Returns:
+        Optional[T]: Parsed output, or None if it could not be parsed
+    """
+    if supports_function_calling(model):
+        chain = create_structured_output_chain(pydantic_object, llm=model, prompt=chat_prompt)
+        
+        try:
+            return await chain.arun(**prompt_input)
+        except (OutputParserException, ValidationError):
+            # In the future, we should probably have some recovery mechanism here (i.e. fix the output with another prompt)
```
pal03377, line 136: Does it make sense to log an error here (also below)?

module_text_llm/module_text_llm/helpers/llm_utils.py
```diff
@@ -0,0 +1,144 @@
...
+    output_parser = PydanticOutputParser(pydantic_object=pydantic_object)
+    system_message_prompt = SystemMessagePromptTemplate.from_template(system_message + "\n{format_instructions}")
+    system_message_prompt.prompt.partial_variables = {"format_instructions": output_parser.get_format_instructions()}
+    system_message_prompt.prompt.input_variables.remove("format_instructions")
+    human_message_prompt = HumanMessagePromptTemplate.from_template(human_message + "\n\nJSON response following the provided schema:")
+    return ChatPromptTemplate.from_messages([system_message_prompt, human_message_prompt])
+
+
+async def predict_and_parse(model: BaseLanguageModel, chat_prompt: ChatPromptTemplate, prompt_input: dict, pydantic_object: Type[T]) -> Optional[T]:
+    """Predicts and parses the output of the model
```
pal03377: With "predict output of the model" you mean that you let the model generate the output, right?

module_text_llm/module_text_llm/helpers/utils.py
```diff
@@ -1,9 +1,14 @@
-import re
 from typing import List, Tuple, Optional
 
+import tiktoken
 from nltk.tokenize import sent_tokenize
 
-from athena.logger import logger
+# This is correct for gpt-4 and chat gpt3.5 but might be different for other models
```
pal03377, line 6: So the token limit might be wrong for other LLMs 🤔 Is there something we can do? It should roughly be correct; maybe we can just lower the limit for other LLMs a bit to be safe?

pal03377 (approved):
Looks mostly good! I only have a few comments where you can decide if they are worth integrating:

module_text_llm/module_text_llm/helpers/llm_utils.py
FelixTJDietrich: As far as I can see there is no property like this, yet. This is a feature of the OpenAI chat models (starting june, but we don't use older models). Some other LLMs support function calling now, maybe they will add it as property someday. Usually, if it is not supported you just get an error during the API call, AFAIK, but I want to catch it earlier.

module_text_llm/module_text_llm/helpers/llm_utils.py
FelixTJDietrich: Yeah, I can change it and word it better

module_text_llm/module_text_llm/helpers/llm_utils.py
FelixTJDietrich: This will happen quite often so I will just add a warning. In the future we can log this to LangSmith etc. to track and improve the error rate

FelixTJDietrich: Actually, nevermind it already gets logged to LangSmith. I won't add anything here

module_text_llm/module_text_llm/helpers/utils.py
FelixTJDietrich: The limits are configurable. Right now I set them to input 3000 and 1000 output tokens. I'm already 96 lower than the limit and I don't think that this will be an issue. In the worst case, we use up a bit more than 3000 actual input tokens and the output token just stop at the limit during generation.

module_text_llm/module_text_llm/generate_suggestions.py
FelixTJDietrich: I removed them now

module_text_llm/module_text_llm/helpers/llm_utils.py
pal03377: Ahh, I see. We don't control the class ourselves

PR #65, opened 2023-07-31
Add evaluation mode running in batch mode

FelixTJDietrich:
### Description

- [x] Refactor BaseInfoHeader and extract the module select and config
- [x] Refactor `mode` route to `data/<data_mode>`
- [x] Add "view mode" selector: `module requests` and `evaluation mode` 
- [x] Define experiment
- [x] Configure module(s) and config(s) to run on
- [x] Conduct Experiment
    - [x] Show config (disabled) + display module type
    - [x] Sync submission views
    - [ ] ~~Rate feedback (correct, relevance, ...)~~ -> Follow-up PR (together with automatic evaluation)
    - [x] Send off data for training for each module (progress bar)
    - [ ] ~~Incremental learning mode with edit feedback! (for finetuning etc.)~~ -> Follow-up PR
    - [x] Batch mode (tutor feedback not editable)
    - [x] Import/Export (progress etc.)
- [x] Fix unreferenced feedback for evaluation data (0-0)

### Steps for Testing

1. Run playground
2. Check if module requests works with the refactored module selector (removed from base info header)
3. Go to evaluation mode
4. Define the experiment you want to conduct 
     - Lots of configuration possible
     - There is a training checkbox where you can optionally enable training data
     - Import/export of the experiment should work
5. Configure modules
     - Now define the configurations you want to use for the experiment
     - Import/export should work (only with the correct exercise type)
6. Conduct experiment
    - Press start -> Progress should be displayed correctly for each module
    - Press "Show config" to look at a modules config (in a dialog)
    - Switch submission to look at (should be in sync with all modules and the tutor feedback)
    - Import/Export basically at any time (might disable it for running modules in the future IDK)

In general, just play around with it and report any issues :)
Repeat for each exercise type (text + programming)

#### Changes to the BaseInfoHeader and the Module Requests mode:
image

#### Define Experiment
image

#### Configure Modules
Screenshot 2023-08-13 at 14 18 43

#### Conduct Experiment
**text** 
image

**programming**
image

PR #80, opened 2023-08-18
Better feedback suggestions in the example module

pal03377:
> **Note**
> To be reviewed & merged after #65

### Motivation and Context

When testing #65, I noticed that the default feedback suggestions for the `module_example` are unreferenced. 

### Description

I wanted some more diversity in the example and therefore added two more feedback suggestions.

### Steps for Testing

You need to wait for #65 to be merged because it seems to fix the detection of unreferenced vs. referenced feedback in the playground.
- Get feedback suggestions from the example module
- Note that there are now three feedback suggestions, two of which are referenced.

### Screenshots

screenshot-2023-08-18_001794


module_example/module_example/__main__.py
```diff
@@ -81,11 +81,43 @@ def suggest_feedback(exercise: Exercise, submission: Submission, module_config:
...
+            title="This is a second suggestion.",
+            description="This is very good!",
+            credits=2.0,
+            file_path="BinarySearch.java",
+            line_start=13,
+            line_end=18,
+            grading_instruction_id=None,
+            meta={}
+        ),
+        Feedback(
```
FelixTJDietrich, line 112: You can also add one for `unreferenced_file` where `line_start` and `line_end` are `None`. Does this make sense? I'm giving feedback like this in my module.

pal03377: Done!

FelixTJDietrich (approved):
LGTM :)

PR #85, opened 2023-08-20
ThemisML: Make startup work on test server

pal03377:
### Motivation and Context

The ThemisML module currently does not start on the test server.

### Description

The reason is that ThemisML tries to access the GPU (indirectly, via pytorch). The pytorch installation for CPU only is difficult to achieve. This PR attempts to install the CPU version still.

I also updated the JavaLexer + JavaParser code from Antlr because it was outdated and I was getting warnings about that.

### Steps for Testing

Start ThemisML on the test server.


module_programming_themisml/Dockerfile
```diff
@@ -22,6 +23,12 @@ RUN poetry config virtualenvs.create true \
     && poetry config virtualenvs.in-project true \
     && poetry install --no-interaction --no-ansi
 
+# re-install pytorch in the CPU version because we don't have a GPU
+RUN poetry source add --priority=supplemental torch-stable https://download.pytorch.org/whl/ # Add the PyTorch repo
+RUN poetry add torch==2.0.0+cpu # Install PyTorch
+# somehow, the torch library needs sympy to be installed manually as well
+RUN poetry add sympy==1.12
```
FelixTJDietrich: The `# comment` part is highlighted like the command in VSCode for me. I don't know if this is working, maybe put the comments on the line before?

```suggestion
# Install PyTorch
RUN poetry add torch==2.0.0+cpu
# somehow, the torch library needs sympy to be installed manually as well
RUN poetry add sympy==1.12
```

FelixTJDietrich (approved):
Other than the one suggestion for the Dockerfile this LGTM, worked as expected on the test server :) 

It ran through, but I probably need to add training submissions to get any suggestions 🤔:
image

module_programming_themisml/Dockerfile
FelixTJDietrich: I don't know if the code suggestion here worked 😅

PR #87, opened 2023-08-23
Docker Production: Add dependencies on Postgres startup

pal03377:
### Motivation and Context

Sometimes, the PostgreSQL database would not start up quick enough with the production setup and therefore, some modules would crash.

### Description

Added a dependency in the production docker-compose file to wait for Postgres to start.

### Steps for Testing

Start the production docker compose file, e.g. by deploying to the test server. Alternative for local setup: `docker compose -f docker-compose.prod.yml up --pull`


FelixTJDietrich (approved):
Deployed successfully, code LGTM

PR #88, opened 2023-08-23
Allow Problem Statement to be null / None

pal03377:
### Motivation and Context

When you don't enter a problem statement for a new text exercise, its value is `null`. This crashes Athena.

### Description

We solve this by allowing the problem statement to be `None` in Athena.

### Steps for Testing

Create a new text exercise on https://ma-schwind.ase.cit.tum.de/. Leave the problem statement empty. Enable automatic assessment. Then, set the due date to something in the future. Athena should not reject the submissions anymore.


FelixTJDietrich (approved):
I checked the code if I could find any other locations where problem_statement could be null/None but could not find any. Code LGTM

PR #89, opened 2023-08-23
Run consumers asynchronously

pal03377:
### Motivation and Context

Sometimes, requests from Artemis for submission sending would time out after 10 seconds because Athena did not answer in time. Athena does not respond with any "real" information, so the response could and should be immediate.

### Description

For the consumer endpoints (which don't return anything), we now run the actual computation from the module asynchronously in the background.

### Steps for Testing

Test the consumers by sending submissions / feedback to Athena. They should respond pretty quickly now.


FelixTJDietrich (approved):
Code LGTM and still works for me on the test server 👍

PR #90, opened 2023-08-27
Allow nullable example solution in text exercises

FelixTJDietrich:
### Motivation and Context

Some text exercises error because they don't have an example solution. For example id: `4279`

### Description

Allows None for the `example_solution` in text exercises.

### Steps for Testing

1. Delete db
2. Run text module + assessment_module_manager
3. Use playground with exported evaluation data 
4. Request a suggestion with a text module for exercise `4279`

### Screenshots

image


pal03377 (approved):
Works! Testes locally

pal03377:
Code also looks good

Issue #81, opened 2023-08-18
Playground: Selecting module without config schema takes long


pal03377:
When you select an assessment module that has no config schema, the playground retries fetching it until the timeout regardless, which makes selecting the assessment module slow:

https://github.com/ls1intum/Athena/assets/9006596/79800982-391c-4e06-9203-1dca0658c565

Optimally, the playground should detect a 404 error from the server and not retry fetching the config.

Ignoring the "loading config" message works. It only feels a bit unnatural.

pal03377:
fixed by #91

PR #91, opened 2023-08-27
Fix long module config loading state

FelixTJDietrich:
### Motivation and Context

The module config shows `Loading...` for a long time when it is not implemented by the module.

### Description

Disable retrying when config loading fails.

### Steps for Testing

1. `module_programming_llm` can be used for testing in the playground
2. Select module in the module requests
3. It should now show `Config options not available, not implemented for this module` much faster.

### Screenshots

image


pal03377:
fixes #81

pal03377 (approved):
Tested locally, it works 👍

PR #92, opened 2023-08-29
Indicate grading instruction with badge

FelixTJDietrich:
### Motivation and Context

@maximiliansoelch pointed out that it is a bit hard to understand that feedback with a proper title are indeed gradin instructions.

### Description

Add a grading instruction badge that shows when a feedback has a `grading_instruction_id`.

### Steps for Testing

1. Use evaluation data or change example data to include a `grading_instrustion_id` in the feedback
2. Preview the feedback in the playground (for instance through `send feedback to Athena`)
3. Verify that the badge is there

### Screenshots

image


pal03377 (approved):
Code looks good & it works!

screenshot-2023-08-30_001873

PR #103, opened 2023-09-07
Programming feedback suggestions integration

pal03377:
### Motivation and Context

We need a few minor changes to support the Artemis integration of programming feedback suggestions, https://github.com/ls1intum/Artemis/pull/7136

### Description

We now pass the Athena secret to programming modules so they can download the programming repositories.

### Steps for Testing

Test https://github.com/ls1intum/Artemis/pull/7136


pal03377:
This PR is contained in #104, so I'll only keep that one

PR #104, opened 2023-09-07
Improvements on ThemisML

pal03377:
### Motivation and Context

While I evaluated ThemisML, I noticed a lot of possible improvements. I implemented them in this PR.

### Description

- Send authentication header to LMS when downloading repos (stored in the Assessment Module Manager, passed to the module through a header, then passed to the LMS as well using FastAPI middleware)
- Small refactorings
- Adapt to line numbers being 1-indexed
- Generate suggestions when receiving feedback instead of when requesting them (pre-generation)
- Completely ignore unreferenced feedback
- Store the number of suggestions generated from a single feedback
- Filter for Java exercises directly for now (ThemisML only supports Java)
- Have a separate cache variable per instance of a `CodeSimilarityComputer` => garbage collect memory more often
- Increase the necessary similarity score from 75% to 95% because it worked better with real data
- Remove async processing because it was making things slower
- Generate feedback suggestions for submissions in batches to avoid memory overflows
- Remove generated overlapping suggestions
- Remove "suspicious" suggestions. A suggestion is suspicious if one of the following happens:
    1. Sometimes, there was a feedback on something banal like a getter, which was actually meant for another method.
        This caused suggestions for almost all the other submissions, which were not helpful.
        We therefore classify a suggestion as "suspicious" if it affects too many other submissions (> 10% and > 2).
    2. However, this would also sometimes classify a suggestion as suspicious if it is actually helpful.
        Therefore, we make a suggestion non-supicious if there are at least 3 other suggestions for the same method.
        This makes a mistake like described above unlikely.
    3. Suggestions are also non-suspicious if they include words that hint at other parts of the code, like "again", "consequential error", "previous", "later", "earlier", "above", "below" and German equivalents of these words.

### Steps for Testing

Test the ThemisML module, e.g. using https://github.com/ls1intum/Artemis/pull/7136 or the playground.


assessment_module_manager/assessment_module_manager/module/request_to_module.py
```diff
@@ -7,6 +7,7 @@
 
 from .module import Module
 from .list_modules import list_modules
+from athena import ExerciseType, authenticate
```
FelixTJDietrich: `authenticate` seems to be unused

athena/athena/helpers/programming/feedback.py
```diff
@@ -1,7 +1,7 @@
 from typing import Optional
 
 
-def format_feedback_text(file_path: Optional[str], line_start: Optional[int] = None, line_end: Optional[int] = None) -> str:
+def format_feedback_title(file_path: Optional[str], line_start: Optional[int] = None, line_end: Optional[int] = None) -> str:
     """Returns a consistent feedback text for an optional file path and line number.
```
FelixTJDietrich: I guess the comment could also be adjusted, i.e. replace text with title, also in the return

module_programming_themisml/module_programming_themisml/feedback_suggestions/feedback_suggestions.py
```diff
@@ -1,87 +1,128 @@
...
+    feedbacks: List[Feedback],
+) -> Iterable[CodeComparisonWithCorrespondingSuggestions]:
+    """Creates code comparisons and corresponding feedback suggestions as a generator."""
+    if len(feedbacks) == 0:
+        return
+    # group feedbacks by file path for faster access
+    logger.debug("Grouping %d feedbacks by file path", len(feedbacks))
+    feedbacks_by_file_path = group_feedbacks_by_file_path(feedbacks)
+    for submission in submissions:
+        for file_path, file_feedbacks in feedbacks_by_file_path.items():
```
FelixTJDietrich, line 64: So basically, when the student can freely choose the file name and it differs from other submissions you cannot copy feedback am I right?

module_programming_themisml/module_programming_themisml/feedback_suggestions/feedback_suggestions.py
```diff
@@ -1,87 +1,128 @@
...
+                continue
+            # get all methods in the file of the submission
+            submission_methods = extract_methods(code)
+            # get all feedbacks that match methods in the submission
+            for s_method in submission_methods:
+                for feedback in file_feedbacks:
+                    if feedback.submission_id == submission.id:
+                        # don't compare feedback with itself
+                        continue
+                    if feedback.meta["method_name"] == s_method.name:
```
FelixTJDietrich, line 82: So the method name is also required to be the same in order to find similar feedback? What if the student can freely choose?

module_programming_themisml/module_programming_themisml/feedback_suggestions/remove_overlapping.py
```diff
@@ -0,0 +1,39 @@
...
+    if feedback1.file_path != feedback2.file_path:
+        return False
+    if feedback1.line_start is None or feedback2.line_start is None:
+        return False
+    if feedback1.line_end is None or feedback2.line_end is None:
+        return False
+    if feedback1.line_start > feedback2.line_end:
+        return False
+    if feedback2.line_start > feedback1.line_end:
+        return False
```
FelixTJDietrich: Equality `==` does not count as overlapping? If it counts then `line_end` is None should also check `feedback1.line_start == feedback2.line_start`

module_programming_themisml/module_programming_themisml/feedback_suggestions/remove_overlapping.py
```diff
@@ -0,0 +1,39 @@
...
+    return True
+
+
+def filter_overlapping_suggestions(suggestions: List[Feedback]) -> List[Feedback]:
+    """
+    Filters out overlapping suggestions we don't want to suggest to tutors.
+    suggestions: List of suggestions to filter
+    """
```
FelixTJDietrich: You could use a more standardized/consistent docstring format. I'm using for example: [autoDocstring - Python Docstring Generator](https://marketplace.visualstudio.com/items?itemName=njpwerner.autodocstring)

I don't know if it makes a big difference, but maybe the documentation page can pick it up better, there are also some other docstrings that are similarly formatted.

module_programming_themisml/module_programming_themisml/__main__.py
```diff
@@ -30,53 +30,74 @@ def select_submission(exercise: Exercise, submissions: List[Submission]) -> Subm
...
+        feedback.meta["n_feedback_suggestions"] = len([f for f in feedback_suggestions if f.meta["original_feedback_id"] == feedback.id])
+        # store the information on the suggestions as well for quicker access later
+        for suggestion in feedback_suggestions:
+            if suggestion.meta["original_feedback_id"] == feedback.id:
+                suggestion.meta["n_feedback_suggestions"] = feedback.meta["n_feedback_suggestions"]
+
+    # save to database
+    store_feedback_suggestions(feedback_suggestions)  # type: ignore
+    for feedback in feedbacks:
         store_feedback(feedback)
```
FelixTJDietrich, line 78: This calls `db.commit()` for each feedback, idk if this is a problem, maybe we should add a helper similar to `store_feedback_suggestions`.

FelixTJDietrich (dismissed):
Code looks good overall :) Just some minor stuff and some questions


I also tested it on the test server and it worked, I guess. Although I thought this result was kind of strange: 🤔 
image
image

pal03377:
I'm merging this now to simplify further PRs. I will fix potential issues that are left in future PRs.

PR #105, opened 2023-09-25
Add environment vars to change module urls

maximiliansoelch:
### Motivation and Context

Kubernetes does not like `_` in the urls.

### Description

Add option to customize the module urls via an environment variable.


FelixTJDietrich:
@pal03377 wasn't this already possible by adjusting the `.ini` file? E.g. `modules.docker.ini` do we need to do it over the environment?

If this is necessary, then the code LGTM but I'm a bit unsure

pal03377:
Alternative solution: #151

pal03377:
Closing this in favor of #151. If #151 does not solve the underlying problem, feel free to re-open :)

maximiliansoelch:
#151 does not completely solve the Kubernetes issue, as we need to address the service instead of the container (e.g., `"http://module-example-service:5001"`).
In general, I think you could name the Kubernetes services to make it work, but I think this is not the best practice.

Further, I like the optional flexibility of overwriting the default module URLs.

FelixTJDietrich (approved):
Alright, in that case this seems fine to me 👍

pal03377:
In this case, I also think that makes sense 👍

pal03377 (approved):

PR #126, opened 2023-10-09
Poetry dependency updates

pal03377:
### Motivation and Context

We want to maintain the poetry packages up to date.

### Description

Update Python packages from all modules, the assessment module manager, and the log viewer.

### Steps for Testing

Test the modules and assessment module manager like normal.


FelixTJDietrich (approved):
I resolved the conflict with the `module_programming_llm` merge. Tested it locally and everything seems to work at least for my modules and the example module. 

I did not test CoFee and ThemisML. Btw. for the VSCode ThemisML terminal seems to be faulty, it does not open the right virtual environment. But maybe that's only on my end 🤔

PR #127, opened 2023-10-09
Fix Docker login (and Test Server Deployment)

pal03377:
### Motivation and Context

The Docker image upload currently [does not work](https://github.com/ls1intum/Athena/actions/runs/6462491466/job/17544195812?pr=125). This also breaks test server deployments.

### Description

Update the way we authenticate against GitHub's package registry using another token, following https://github.com/docker/login-action#github-container-registry

### Steps for Testing

Test the Docker image build & upload.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -68,7 +68,7 @@ jobs:
         with:
           registry: ghcr.io
           username: ${{ github.repository_owner }}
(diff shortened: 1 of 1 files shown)


pal03377:
Because the change is just one simple line and critical for further development, I will merge this right away.

PR #150, opened 2023-10-22
Remove all unused images on cleanup, not just dangling ones

pal03377:
### Motivation and Context

The test server disk space got full because non-dangling Docker images were not removed.

### Description

Change the existing Docker cleanup (after TS deployment) to also remove non-dangling images.

### Steps for Testing

Deploy to the test server. Then, observe how old images are removed and there won't be a disk space problem for a longer time span.

### Screenshots

I manually executed the cleanup command to test:
screenshot-2023-10-22_002031

### DIFF
--- a/athena-docker.sh
+++ b/athena-docker.sh
@@ -101,7 +101,7 @@ function all_logs {
 }
 
 function cleanup {
-  docker image prune -f
+  docker image prune -a -f # remove all unused images
 }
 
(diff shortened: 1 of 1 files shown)


pal03377:
I'm merging this directly because it works and it's very valuable to have right away

FelixTJDietrich:
Thanks @pal03377! :)

PR #151, opened 2023-10-22
Hostnames with dashes

pal03377:
### Motivation and Context

Hostnames using underscores should technically not be allowed. We currently cannot deploy to Kubernetes because of this (see #105 for the quickfix).

### Description

Use dashes in the hostnames by default.

### Steps for Testing

Test using the assessment modules in Athena on the test server. They should connect to the assessment module manager properly.


pal03377:
@maximiliansoelch Do you think this solves the Kubernetes deployment problems? In this case, this also closes #105

FelixTJDietrich (approved):
Code LGTM I think this is also a good solution

PR #153, opened 2023-10-22
ThemisML Python upgrade

pal03377:
### Motivation and Context

The Python version of the ThemisML Docker container is not up to date.

### Description

Update the Python version to 3.11.

### Steps for Testing

Run ThemisML.


pal03377:
Merging this directly as it is only a dependency update.

PR #144, opened 2023-10-19
Support Structured Grading Instructions

FelixTJDietrich:
### Motivation and Context

The `module_text_llm` and `module_programming_llm` need the structured grading instructions and we want to send it to Athena as a JSON object.

### Description

- [x] Update export scripts
- [ ] SGIs for text exercises
- [ ] SGIs for programming exercises
- [ ] Update playground UI
- [ ] Update LLM modules

PR #154, opened 2023-10-22
Add support for grading criteria

pal03377:
### Motivation and Context

We need to access the structured grading criteria for the LLM modules.

### Description

Add support for sending structured grading criteria (= collections of structured grading instructions, not to be confused with the unstructured grading instructions as a string on the exercise directly).

### Steps for Testing

- Test that the evaluation data export still works as expected and also exports the structured grading criteria.
- Test via the playground that each one text and one programming module still work and properly receive the sent data.

---
This PR is very similar to #144 because I failed to communicate that I'm working on this 😬  We'll merge our efforts here.


athena/athena/models/db_feedback.py
```diff
@@ -12,7 +12,7 @@ class DBFeedback(Model):
     title = Column(String)
     description = Column(String)
     credits = Column(Float, nullable=False)
-    grading_instruction_id = Column(BigInteger)
+    structured_grading_instruction = Column(JSON)
```
FelixTJDietrich: I'm fine with this change but I think we can just keep `grading_instruction_id`, this is also more compatible with Artemis. 

For my modules it is not important to have it as JSON

athena/athena/models/db_programming_feedback.py
```diff
@@ -39,5 +39,5 @@ def get_referenced_code(self) -> str:
                 return code
             if self.line_end is None:
                 raise ValueError("Unexpected: line_start is not None, but line_end is None (only line_start given)")
-            # Only return the lines between line_start and line_end. Line numbers are 1-based.
-            return "\n".join(code.split("\n")[self.line_start - 1 : self.line_end])
+            # Only return the lines between line_start and line_end. Line numbers are 0-based.
+            return "\n".join(code.split("\n")[self.line_start : self.line_end + 1])
```
FelixTJDietrich, line 43: Works for me :+1: will have to adjust it in the programming LLM module as well, I saw that you changed something in `module_text_llm` will have to check if this is correct.

athena/athena/schemas/feedback.py
```diff
@@ -14,7 +15,7 @@ class Feedback(Schema, ABC):
                              example="Your solution is correct.")
     credits: float = Field(0.0, description="The number of points that the student received for this feedback.",
                            example=1.0)
-    grading_instruction_id: Optional[int] = Field(None, description="The id of the grading instruction that this feedback belongs to.", example=1)
+    structured_grading_instruction: Optional[StructuredGradingInstruction] = Field(None, description="The id of the structured grading instruction that this feedback belongs to.", example=1)
```
FelixTJDietrich: Here the description and example are wrong but maybe we might want to change it back to be `grading_instruction_id` or `structured_grading_instruction_id`

athena/athena/schemas/grading_criterion.py
```diff
@@ -0,0 +1,26 @@
...
+    grading_scale: str = Field(description="The grading outcome for this instruction.", example="Weak example", default="")
+    instruction_description: str = Field(description="Description of how to use this grading instruction.", example="Some instructions", default="")
+    feedback: str = Field(description="Description of the feedback to provide.", example="Nicely done!", default="")
+    usage_count: int = Field(ge=0, description="The usage limit for this structured grading instruction. 0 means unlimited.", example=3, default=0)
+
+
+class GradingCriterion(Schema, ABC):
+    """A structured grading criterion for assessing an exercise."""
+    id: int = Field(example=1)
+    title: Optional[str] = Field(None, example="Some instructions")
```
FelixTJDietrich, line 23: Maybe add a `description` here for completeness also with the `structured_grading_instructions`

`title` -> `description="Title of the criterion grouping multiple structured grading instructions of different grading scales.", example="Assessment of Correctness"` 

`structured_grading_instructions` -> `description="List of structured grading instructions for the same criterion."`

But idk.

module_example/module_example/__main__.py
```diff
@@ -91,7 +91,7 @@ def suggest_feedback(exercise: Exercise, submission: Submission, module_config:
             file_path="BinarySearch.java",
             line_start=8,
             line_end=9,
-            grading_instruction_id=None,
+            structured_grading_instruction=None,
```
FelixTJDietrich: Actually, I would now prefer keeping `grading_instruction_id` otherwise the export gets more complicated again ^^

module_text_llm/module_text_llm/helpers/utils.py
```diff
@@ -44,17 +44,17 @@ def get_index_range_from_line_range(line_start: Optional[int], line_end: Optiona
...
     if line_start > line_end:
         line_start, line_end = line_end, line_start
 
     sentence_spans = get_sentence_spans(content)
-    line_start_index = int(line_start) - 1
+    line_start_index = int(line_start)
     line_start_index = min(max(line_start_index, 0), len(sentence_spans) - 1)
 
-    line_end_index = int(line_end) - 1
+    line_end_index = int(line_end)
```
FelixTJDietrich: I think you meant to change this for the `module_programming_llm`, but I think it is fine for consistency reasons. We also need to change it in `module_programming_llm` and update the add line numbers helper within the prompt so the LLM knows that it starts with `0`.

playground/src/components/details/editor/inline_feedback.tsx
```diff
@@ -199,9 +201,10 @@ export default function InlineFeedback({
                   Suggestion
                 </span>
               )}
-              {feedback.grading_instruction_id && (
+              {feedback.structured_grading_instruction && (
```
FelixTJDietrich: Now I see why you wanted it as json. But I think the ID would still be better.

playground/src/components/details/editor/inline_feedback.tsx
```diff
@@ -199,9 +201,10 @@ export default function InlineFeedback({
...
                 <span className="text-xs text-orange-800 rounded-full px-2 py-0.5 bg-orange-100">
-                  Grading Instruction
+                  Grading Instruction:
+                  {feedback.structured_grading_instruction.feedback}
```
FelixTJDietrich: This would render strangely, it would end up being too long, I think.

Also right now I'm automatically filling `description` and the `title` with the grading instructions feedback description and grading criterion title. I think if at all it should be ` Grading Instruction {feedback.structured_grading_instruction.id}` or just `Grading Instruction`, it's just a badge.

FelixTJDietrich (changes requested):
Thanks for adding this! There are some things that need to be adjusted but great work so far :)

athena/athena/models/db_programming_feedback.py
FelixTJDietrich: I did it

module_text_llm/module_text_llm/helpers/utils.py
FelixTJDietrich: I changed it 👍

FelixTJDietrich (approved):
Looks good now, will adjust the follow-up PRs tomorrow 👍

PR #161, opened 2023-10-26
Update langchain

FelixTJDietrich:
### Motivation and Context

closes #158 
closes #159

### Description

Update dependencies

### Steps for Testing

Check if LLM modules are still generating feedback


FelixTJDietrich:
@pal03377 the top-level `poetry.lock` also needs to be updated I think 🤔 Otherwise the security issues don't go away AFAIK. 

I was trying to update it using `poetry lock` (IDK if this is the correct way) but got:
image

I tried changing the python version in the top-level `poetry.toml` file but it did not help

FelixTJDietrich:
@pal03377 Also another point. Stephan asked me to fix the dependencies of the Playground to a fixed version so without a `^`. I thought this is not really necessary since that's the job of the lock files anyways but I can see some benefits.

Do you think we should also fix the versions of the python dependencies?

pal03377:
@FelixTJDietrich I fixed the lockfile issue with #165. I suggest we merge that one first and then update this PR.

FelixTJDietrich:
Tested locally will merge as soon as the checks pass

PR #165, opened 2023-10-26
Set all Python versions to 3.11

pal03377:
### Motivation and Context

Incompatible Python version requirements caused issues with running `poetry lock` in the project root.

### Description

Fix Python version to `3.11.*`.

### Steps for Testing

Test if the modules still run and that the CI runs through.


FelixTJDietrich (changes requested):
Looks good to me but Prospector is failing. Will also test once it's passing.

pal03377:
I tested the example module and ThemisML and they both work

FelixTJDietrich (approved):
Both llm modules are still working, will merge it right away

PR #170, opened 2023-11-01
Update prompts and general improvements and fixes

FelixTJDietrich:
### Motivation and Context

Some fixes and improvements of #84 

### Description

- Fix model temperatures to be more deterministic (this is what we want)
- Update all prompts with best prompts from the thesis
- Fix export of results to include all necessary ids
- Fix rendering issues when opening the same submissions with Monaco editor
- Fix experiment changes do not change experiment id 

### Steps for Testing

- Run an evaluation in the playground with either LLM module
- It should still work and the exported results should contain the experiment, moduleConfiguration, and run ids to identify them

### DIFF
--- a/module_programming_llm/module_programming_llm/helpers/models/replicate.py
+++ b/module_programming_llm/module_programming_llm/helpers/models/replicate.py
@@ -38,6 +38,7 @@ if os.environ.get("REPLICATE_API_TOKEN"):  # If Replicate is available
     available_models = {
         name: Replicate(
             model=model,
+            input={ "temperature": 0.01 }
         )
         for name, model in replicate_models.items()
     }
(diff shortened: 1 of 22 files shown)


module_programming_llm/module_programming_llm/helpers/models/replicate.py
```diff
@@ -38,6 +38,7 @@
     available_models = {
         name: Replicate(
             model=model,
+            input={ "temperature": 0.01 }
```
pal03377, line 41: Why not use a temperature of 0 here?

pal03377 (approved):
The code changes make sense to me. I also tested the PR on the test server. I just have a small question to answer, then I think we can merge this 👍

module_programming_llm/module_programming_llm/helpers/models/replicate.py
FelixTJDietrich: This is a detail from the LLaMA 2 models, their minimum temperature is `0.01`:
See https://replicate.com/meta/llama-2-70b-chat

PR #171, opened 2023-11-05
Add support for manual evaluation

FelixTJDietrich:
### Motivation and Context

Reworked manual ratings from #84 for adding manual evaluation to the playground

### Description

- Add manual ratings to the evaluation mode (accept/reject buttons)
- Add export of manual ratings separately from the results (same button but two files, to make it easier for multiple annotators)
- Add import for manual ratings which can be imported after the results (same button, and we automatically order multiple input files)

### Steps for Testing

1. Run eval mode for a text module
2. Annotate some suggestions and press export, it should export two files (results and manualRatings)
3. Cancel experiment (in module configuration and define the module configuration again)
4. Try out a combination of imports, i.e. first results then manual ratings, both at once -> success | no results then manual ratings or wrong manual ratings (of a different run for example) -> error alert 
5. Repeat for a programming module

#### Inline Feedback for Text Modules
image

#### Inline Feedback for Programming Modules
image


pal03377 (approved):
I tested the PR on the test server, using both exercise types and multiple modules. Everything seems to work as intended! The code also looks good.

PR #174, opened 2023-11-05
Add automatic evaluation with LLM-as-a-Judge, LangSmith export, and SGI evaluation

### LINKED ISSUE #36: Initial Evaluation Mode Roadmap
## Motivation

We want to evaluate the performance of modules for research purposes to better compare and iterate. This is a roadmap for the initial evaluation mode that at least @FelixTJDietrich wants to use for his Master's thesis.

## Roadmap

- [x] #35 
- [x] #37
- [x] #44 
- [x] #52 
- [x] #56 
- [x] #61 
- [x] #64 
- [x] #70 
    - [x] Settle on evaluation data to use for final evaluation
- [x] #65 
- [x] #92
- [x] #160
- [x] #166
- [x] #170 
- [ ] #171 
- [ ] #173

image

FelixTJDietrich:
### Motivation and Context

We want an option to systematically evaluate the modules without manual labor. 

closes #84 
closes #36 

### Description

- [x] Add `/evaluation` endpoint to Athena for returning arbitrary evaluation results 
- [x] Add example of how this `/evaluation` endpoint would like to `module_example` (for programming exercises)
- [x] Add this new module request to the Playground
- [x] Run evaluation for all healthy modules of the same type that support evaluation in evaluation mode
- [x] Display progress of automatic evaluation (and move "Start Generating" button)
- [x] Add LLM-as-a-judge for text modules
- [x] Disable LLM-as-a-judge using environment variable
- [x] Add token usage and response times in evaluation response (coming from LangSmith)
- [x] Add eval code for SGI evaluation

**Note:** I omitted the "module_text_llm estimated Accepted with Minor modifications" text from the UI since it is kind of ugly to add and does not give immediate benefit. In the future one might make automatic results available for the inline feedback view but this depends also on future experiment designs. 

### Steps for Testing

1. Run an evaluation with `module_text_llm`
2. After everything is done click export, it should export the `automatic_evaluation` now as a separate file
3. File should contain `LLM-as-a-judge`, `llm_statistics` (from LangSmith, I added my key to the env), and `feedback_statistics`
4. Cancel experiment
5. Import results and automatic evaluation and check if it works

### Screenshots

**Docs:**
image

#### Not Started
Screenshot 2023-11-07 at 16 36 26

#### Sent Training Feedback
Screenshot 2023-11-07 at 16 36 31

#### Generated Suggestions
Screenshot 2023-11-07 at 16 38 37

#### Finished
Screenshot 2023-11-07 at 16 38 42


athena/athena/endpoints.py
```diff
@@ -358,4 +358,65 @@ def config_schema_provider(cls: Type[C]) -> Type[C]:
...
+            true_feedbacks: List[feedback_type], 
+            predicted_feedbacks: List[feedback_type],
+        ):
+        # Retrieve existing metadata for the exercise, submission and feedback
+        exercise.meta.update(get_stored_exercise_meta(exercise) or {})
+        submission.meta.update(get_stored_submission_meta(submission) or {})
+        for feedback in true_feedbacks:
+            feedback.meta.update(get_stored_feedback_meta(feedback) or {})
+        for feedback in predicted_feedbacks:
+            feedback.meta.update(get_stored_feedback_meta(feedback) or {})
```
pal03377:
```suggestion
        for feedback in true_feedbacks + predicted_feedbacks:
            feedback.meta.update(get_stored_feedback_meta(feedback) or {})
```

module_example/module_example/__main__.py
```diff
@@ -139,5 +140,30 @@ def suggest_feedback(exercise: Exercise, submission: Submission, module_config:
     ]
 
 
+# Only if it makes sense for a module (Optional)
```
pal03377, line 143: Very good that you included an eval endpoint to the example module! This will help future developers in their understanding, I believe.

module_text_llm/module_text_llm/evaluation.py
```diff
@@ -0,0 +1,116 @@
...
+    # Count SGIs in suggested feedbacks
+    for feedback in predicted_feedbacks:
+        if feedback.structured_grading_instruction_id:
+            suggestions_with_grading_instructions.append(feedback)
+            suggested_sgi_usage[feedback.structured_grading_instruction_id] += 1
+
+    actual_feedback_with_grading_instructions_count = len(
+        actual_feedback_with_grading_instructions)
+    suggestions_with_grading_instructions_count = len(
+        suggestions_with_grading_instructions)
```
pal03377: I think you can directly write out the `len(...)` twice. This looks like a micro-optimization to me.

module_text_llm/module_text_llm/helpers/utils.py
```diff
@@ -92,3 +92,28 @@ def get_index_range_from_line_range(line_start: Optional[int], line_end: Optiona
     line_end_index = min(max(int(line_end), 0), len(sentence_spans) - 1)
     
     return sentence_spans[line_start_index][0], sentence_spans[line_end_index][1]
+
+
+def get_line_range_from_index_range(index_start: Optional[int], index_end: Optional[int], content: str) -> Tuple[Optional[int], Optional[int]]:
```
pal03377, line 97: Do this function and the one before make sense in a more general utils file in the `athena` package? They seem useful.

playground/src/hooks/athena/request_feedback_suggestions.ts
```diff
@@ -40,6 +40,7 @@ export default function useRequestFeedbackSuggestions(
       }
       return response;
     },
+    retry: 3,
```
pal03377: Why do we need this?

pal03377 (commented):
I looked through the code already. I will test everything later as well.

pal03377 (changes requested):
I get no automatic evaluation and no feedback about this at all when trying it on the test server :(

screenshot-2023-11-08_002086

pal03377:
Also, the `automatic_evaluation` file did not contain the `LLM-as-a-judge`, `llm_statistics`, and `feedback_statistics` keys:
```json
{
  "type": "automaticEvaluation",
  "runId": "f91cc124-8d30-4a50-a0c0-6c039a1738a8",
  "experimentId": "fd22a762-b06f-4b52-98e6-b9227e27dacb",
  "moduleConfigurationId": "145837b8-5ed0-42f6-8709-aad12450a7d6",
  "submissionsWithAutomaticEvaluation": {
    "204": {}
  }
}
```

module_text_llm/module_text_llm/helpers/utils.py
FelixTJDietrich: It is rather module specific I think, but generally could see it also being part of the Athena package. I would not move it now because in `get_sentence_spans` I'm using `nltk`'s sentence tokenization which I need to download in `__main__`:
```
if __name__ == "__main__":
    nltk.download("punkt")`
```

playground/src/hooks/athena/request_feedback_suggestions.ts
FelixTJDietrich: Locally I did not need this, but when I sent feedback for more than `50` training submissions in the evaluation mode on the test server I did get stuck some time. With this it works with `500` training submissions without fail, if it still fails you need to export -> cancel -> import to recover for now.

But you have a point. I'm also using this in "Module Requests" and there we should not retry

module_text_llm/module_text_llm/evaluation.py
FelixTJDietrich: Yeah, I'm no longer misusing `suggestions_with_grading_instructions` as `unmatched_ suggestions_with_grading_instructions` now and can use `len(` now below

playground/src/hooks/athena/request_feedback_suggestions.ts
FelixTJDietrich: Moved it to the evaluation mode hook 👍

FelixTJDietrich:
Seems like I did not validate the `structured_grading_instruction_id`, it should be fine now.

FelixTJDietrich:
Works for me on the test server now :)
image
image

FelixTJDietrich:
I also added now some docs:
image

pal03377 (approved):
The exported files look like expected. I also cannot find any issues (neither in the code nor in my tests) ✅

PR #175, opened 2023-11-06
Fix OpenAI API key error when unset

FelixTJDietrich:
### Motivation and Context

When `OPENAI_API_KEY` is unset LangChain throws an error.

image (2)

### Description

Set `OPENAI_API_KEY=""` programatically so it does not throw.

### Steps for Testing

Remove `OPENAI_API_KEY` from environment and run a LLM module feedback generation request.


FelixTJDietrich:
This works as expected. I reproduced the error and tested it. The added lines prevent the exception. I will merge it now since testing it is kind of complicated and I tested it extensively.

PR #234, opened 2024-03-08
`Programming LLM`: Automatic Feedback for Programming Exercises

undernagruzez:
### Motivation and Context
This PR updates the feedback_suggestion endpoint in the programming_llm module to include an is_graded parameter. This addition allows distinguishing between graded and non-graded feedback. It also introduces prompts for summarizing source code and generating non-graded feedback, intended for student use.

### Description
Changes include:

Addition of is_graded parameter to feedback_suggestion endpoint.
New prompts for source code summaries.
New functionality for generating non-graded feedback suggestions.

### Steps for Testing
In conjunction with a future Artemis PR


module_programming_llm/module_programming_llm/__main__.py
```diff
@@ -1,35 +1,70 @@
...
+        submission.id,
+        exercise.id,
+    )
 
 
 @feedback_provider
-async def suggest_feedback(exercise: Exercise, submission: Submission, module_config: Configuration) -> List[Feedback]:
-    logger.info("suggest_feedback: Suggestions for submission %d of exercise %d were requested", submission.id, exercise.id)
-    return await generate_suggestions_by_file(exercise, submission, module_config.approach, module_config.debug)
+async def suggest_feedback(exercise: Exercise, submission: Submission, is_graded: bool, module_config: Configuration) -> \
```
FelixTJDietrich: Just randomly checked out this PR, why not put the `is_graded` property in `Exercise`? Maybe even optional to not break stuff.

undernagruzez: Do I understand you correctly that you're suggesting adding the is_graded property to the Exercise entity? is_graded is rather a parameter of a request, not of an exercise. We can generate non graded feedback for an exercise at first, and then, after the deadline, graded.

FelixTJDietrich: I see 🤔 but since `exercise` is also sent with each request it would still be possible to generate both. Anyways, your way makes more sense if it is less on an exercise level but rather on the feedback level 👍

athena/athena/endpoints.py
```diff
@@ -270,7 +270,9 @@ def feedback_provider(func: Union[
     Callable[[E, S], List[F]],
     Callable[[E, S], Coroutine[Any, Any, List[F]]],
     Callable[[E, S, C], List[F]],
-    Callable[[E, S, C], Coroutine[Any, Any, List[F]]]
+    Callable[[E, S, C], List[F]],
+    Callable[[E, S, C], Coroutine[Any, Any, List[F]]],
+    Callable[[E, S, G, C], List[F]],
```
maximiliansoelch: `Callable[[E, S, C], List[F]],` is there twice

```suggestion
    Callable[[E, S, C], Coroutine[Any, Any, List[F]]],
    Callable[[E, S, G, C], List[F]],
```

maximiliansoelch (commented):
Can we deploy it to the test server and test it in conjunction with your Artemis PRs?

maximiliansoelch (approved):

PR #237, opened 2024-03-24
`Modeling Exercises`: Filter out non-referenceable element IDs

matthiaslehnertum:
### Motivation and Context
For modeling exercises, the translation of an Apollon diagram into the most suitable format for evaluation by an LLM sometimes adds auxiliary elements in the output which do not directly correspond to an element in the original input but are required by the translation-output format.

An example of this behaviour for BPMN diagrams is the introduction of a "process" element which is purely a syntactic requirement but does not offer semantic information about the depicted process:

```XML
<bpmn:process id="id_78" isExecutable="false">
      <bpmn:laneSet id="id_86" name="Loan Applicant"/>
      <bpmn:startEvent id="id_2" name="">
          <bpmn:outgoing>id_40</outgoing>
      </bpmn:startEvent>
      <bpmn:task id="id_4" name="Send credit request">
          <bpmn:incoming>id_40</incoming>
          <bpmn:outgoing>id_42</outgoing>
      </bpmn:task>
</bpmn:process>
```

The problem here is, that when we ask the LLM to reference element IDs a given feedback item applies to, the LLM sometimes references elements like the "process" which do not have a direct correspondence in the original input diagram. This causes issues as no element with the referenced ID can be found when trying to attach the feedback items in the exercise assessment view in Artemis. This PR introduces filtering logic to omit references that are not present in the original input diagram but only part of the intermediary format.

### Description
This change introduces the function `def filter_ids_for_model(ids: List[str], model: dict) -> List[str]:` in `module_modelling_llm/module_modelling_llm/generate_suggestions.py` which is used to filter non-referenceable element IDs.

```Python
feedbacks.append(Feedback(
          exercise_id=exercise.id,
          submission_id=submission.id,
          title=feedback.title,
          description=feedback.description,
          element_ids=filter_ids_for_model(element_ids, submission_diagram),
          credits=feedback.credits,
          structured_grading_instruction_id=grading_instruction_id,
          meta={}
))
```

### Steps for Testing
When requesting feedback for a modeling exercise, no element IDs should be contained in the "elementIds" array of the JSON response to were not present in the assessed diagram.


marlon-luca-bu (approved):
Code looks good, array didn't contain any element IDs

PR #280, opened 2024-05-20
Fix function calling, more verbose logging, OpenAI upgrade

undernagruzez:
### Motivation and Context
Langchain release a new major upgrade. The version in module programming llm was far behind and had to be upgraded. This PR addresses this and also adds a bit better logging.

### Description

What was changed:
1. The way llm chains are built - was updated to use the new features of langchain.
2. The way LLM models are initialized - because of stuff removed from the long chain.
3. A flag to support more verbose logging.
4. Repeated LLM invocations - in case of failures, additional invocations will be issued, up to 3 times.

### Steps for Testing

Just deploy it to the test server and request feedback and suggestions from Athena and Artemis.

PR #284, opened 2024-06-03
Upgrade Playground

undernagruzez:
### Motivation and Context
Playground was not extended to support self-learning feedback generation, although this feature exists.
Also, this PR https://github.com/ls1intum/Athena/pull/280 changed some internals and changed Authentication.
Playground was not affected. This PR fixes that.

### Description
1. Renamed all URL fields into uri fields to stay consistent with Artemis.
2. Added deployment information about Playground to enable support for new Authentication.
3. Updated nodejs version for docker images.
4. Added lms(client) URL field to the playground header.
5. Chaned athena fetcher to modify the authorization headers in Playground client.
6. Chaned Playground server to forward these headers.
7. Removed unnecessary yarn lock file.
### Steps for Testing
Deploy the testing setup on the ts or locally. Submit new self-learning-feedback requests and old graded requests. Verify everything works.
For local, the key is empty. For athena ts3, the key is here: https://confluence.ase.in.tum.de/pages/viewpage.action?pageId=25252245

### Screenshots
image


athena/athena/models/model.py
```diff
@@ -1,9 +1,17 @@
 import importlib
 
 from pydantic import BaseModel
+from sqlalchemy import Column, String, UniqueConstraint
 
 
 class Model:
+
+    lms_url = Column(String, index=True, nullable=False)
```
FelixTJDietrich: Why not use `repository_uri` from `programming_submission.py`?

undernagruzez: repository urls can differ: we can use gitlab, local vc etc

FelixTJDietrich: Why can‘t you just send the correct url to Athena then?

undernagruzez: It will be sent: https://github.com/ls1intum/Artemis/pull/8635
Or what do you mean?

FelixTJDietrich: But why? Why set it in the header?
Artemis sends the programming submission containing a repository_url (or uri idk). Why is the server not set there?

undernagruzez: > But why? Why set it in the header?
> Artemis sends the programming submission containing a repository_url (or uri idk). Why is the server not set there?

I’m not sure I understand your suggestion. Are you proposing to set the LMS URL in the DTOs sent by the LMS, instead of setting it in the headers? If you mean extracting the LMS URL from the programming exercise properties, that won’t work, as we cannot extract this data for text and modeling exercises.

FelixTJDietrich: I guess I get it now 🤔 
Paul and me just assumed that we have one secret for the whole Athena server. But I guess your problem is more that you'd have submission ids of different LMSs colliding etc.

undernagruzez: Yes exactly, since we plan to use this only testserver(and potentially the prod servers) with multiple Artemis instances, we need this change, but it comes from another PR, and this is just a stacked PR.

DominikRemo (changes requested):
Tested locally and on ts. I found that I get the error **Failed to get config options from Athena**, which I do not get on develop. So I couldn't really test it.
Bildschirmfoto 2024-06-23 um 18 03 48

FelixTJDietrich:
@DominikRemo you probably did not use the Athena Testserver Secret from confluence

undernagruzez:
@FelixTJDietrich @DominikRemo there was actually an issue with the dockerfile, it didn't want to move a file in the container...

DominikRemo (changes requested):
I no longer get the same error message from before, but now I get another error message:
Bildschirmfoto 2024-06-24 um 14 16 16
I tried with the Module Requests mode.
Chose the Module _module_text_cofee_
And then tried to send submissions (does not matter what exercise)
It still works on develop, but not on this branch. Do I need to configure something differently or is it a fault on your site?

undernagruzez:
> I no longer get the same error message from before, but now I get another error message: Bildschirmfoto 2024-06-24 um 14 16 16 I tried with the Module Requests mode. Chose the Module _module_text_cofee_ And then tried to send submissions (does not matter what exercise) It still works on develop, but not on this branch. Do I need to configure something differently or is it a fault on your site?

Did you use the key from the confluence page: https://confluence.ase.in.tum.de/pages/viewpage.action?pageId=25252245?
You must set this field: 
image

DominikRemo (approved):
I tested it locally.
Anyway, I had another look at it on the test server and it works there. I'll probably need to reconfigure my local setup, but it LGTM on test server

DominikRemo (dismissed):
Tested locally and on Athena ts3.

Tested graded and non graded feedback for programming_llm and tested text_cofee and text_llm successfully.

Thank you very much for the fixes!

EneaGore (dismissed):
Looking good to me in the test server.

laadvo (dismissed):
Code looks good and the playground is working on the test server again. Thanks so much for the quick fix! :slightly_smiling_face:

assessment_module_manager/deployments.docker.ini
```diff
@@ -1,2 +1,8 @@
 [schwind]
-url = https://ma-schwind.ase.cit.tum.de
\ No newline at end of file
+url = https://ma-schwind.ase.cit.tum.de
```
FelixTJDietrich, line 2: Is this correct? I tried to resolve the conflicts

assessment_module_manager/deployments.ini
```diff
@@ -1,4 +1,7 @@
 # for local development only
 
 [local]
-url = http://localhost:8080
\ No newline at end of file
+url = http://localhost:8080
```
FelixTJDietrich, line 4: Is this correct? I tried to resolve the conflicts

assessment_module_manager/assessment_module_manager/endpoints/health_endpoint.py
```diff
@@ -37,7 +37,8 @@ class HealthResponse(BaseModel):
                     "url": "http://localhost:5001",
                     "type": "programming",
                     "healthy": True,
-                    "supportsEvaluation": True
+                    "supportsEvaluation": True,
+                    "supportsNonGradedFeedbackRequests": True
```
FelixTJDietrich: I think we would also want to have:
`supportsGradedFeedbackRequests`

env_example/assessment_module_manager.env
```diff
@@ -12,4 +12,4 @@ MODULE_PROGRAMMING_THEMISML_SECRET=12345abcdef
 # LMS Deployments                                          #
 ################################################################
 # the deployment name should correspond to the name in deployments.ini
-LMS_DEPLOYMENT_NAME_SECRET=12345abcdef
\ No newline at end of file
+LMS_DEPLOYMENT_NAME_SECRET=12345abcdef
```
FelixTJDietrich, line 15: I think I accidentally removed a new line

assessment_module_manager/modules.docker.ini
```diff
@@ -2,33 +2,40 @@
 url = http://module-example:5001
 type = programming
 supports_evaluation = true
+supports_non_graded_feedback_requests = false
```
FelixTJDietrich, line 5: Can we automatically detect this? :) 
Also for evaluation.

module_modeling_llm/module_modeling_llm/__main__.py
```diff
@@ -28,7 +28,7 @@ def process_incoming_feedback(exercise: Exercise, submission: Submission, feedba
 
 
 @feedback_provider
-async def suggest_feedback(exercise: Exercise, submission: Submission, module_config: Configuration) -> List[Feedback]:
+async def suggest_feedback(exercise: Exercise, submission: Submission, is_graded: bool, module_config: Configuration) -> List[Feedback]:
```
FelixTJDietrich, line 31: I still think it would be cool to have:
`@non_graded_feedback_provider` and `@graded_feedback_provider` as decorators

It would make it clearer, but this is probably out of scope for the PR

playground/src/components/view_mode/evaluation_mode/conduct_experiment/module_experiment_progress.tsx
```diff
@@ -22,7 +22,7 @@ export default function ModuleExperimentProgress({
       "notStarted",
       "sendingSubmissions",
       "sendingTrainingFeedbacks",
-      "generatingFeedbackSuggestions",
+      "generatingGradedFeedbackSuggestions",
```
FelixTJDietrich, line 25: I guess we would also have the graded vs non-graded destination for the evaluation mode. Like a checkmark in the beginning and then later you can also only select modules that support it.

playground/src/components/view_mode/module_requests/request_non_graded_feedback_suggestions.tsx
```diff
@@ -0,0 +1,137 @@
...
+
+    const feedbacks = response.data;
+    return (
+      submission && (
+        <Disclosure
+          title="Submission with Non-Graded Feedback Suggestions"
+          openedInitially
+          className={{ root: "ml-2" }}
+        >
+          <SubmissionDetail identifier="suggestions" submission={submission} feedbacks={feedbacks} />
```
FelixTJDietrich, line 65: Are we rendering non-graded feedback differently? 
As a follow-up issue we could think about if the schema of `NonGradedFeedback` and  `GradedFeedback` should be different 🤔
The `NonGradedFeedback` does not have credits and structured grading instructions and probably needs some other attributes about positivity, priority, etc.

Can you create a follow-up issue for that. This would also then be returned by the `@non_graded_feedback_provider` and `@graded_feedback_provicer` respectively.

playground/src/hooks/batch_module_experiment.ts
```diff
@@ -297,8 +297,8 @@ export default function useBatchModuleExperiment(experiment: Experiment, moduleC
   };
 
   // 3. Generate feedback suggestions
-  const stepGenerateFeedbackSuggestions = async () => {
-    setProcessingStep("generatingFeedbackSuggestions");
+  const stepGenerateGradedFeedbackSuggestions = async () => {
```
FelixTJDietrich, line 300: Can you also make it work for non-graded feedback? :) 

But I think we have to do this more as a follow-up issue. Can you create it? I.e. `Adjust evaluation mode to support non-graded feedback`. This would also include displaying it correctly

FelixTJDietrich (changes requested):

playground/src/hooks/batch_module_experiment.ts
undernagruzez: I created an issue: https://github.com/ls1intum/Athena/issues/317

playground/src/components/view_mode/module_requests/request_non_graded_feedback_suggestions.tsx
undernagruzez: I will create an issue, but we use the same schema for Athena modules. We can discuss this in dm or in the Athena meeting.
https://github.com/ls1intum/Athena/issues/318

playground/src/components/view_mode/evaluation_mode/conduct_experiment/module_experiment_progress.tsx
undernagruzez: For a follow-up that extends the evaluation mode?

assessment_module_manager/deployments.docker.ini
undernagruzez: Should be correct.

assessment_module_manager/deployments.ini
undernagruzez: same

assessment_module_manager/assessment_module_manager/endpoints/health_endpoint.py
undernagruzez: I thought all modules should have graded feedback suggestions by default. If that is not the case, I can quickly modify this part.

assessment_module_manager/modules.docker.ini
undernagruzez: See the answer below.

module_modeling_llm/module_modeling_llm/__main__.py
undernagruzez: We might decide to use different decorators. Then, we could scan over the main.py and see if the corresponding decorator is within the file to automatically set the flags.

assessment_module_manager/assessment_module_manager/endpoints/health_endpoint.py
FelixTJDietrich: Could be optional in my opinion. Maybe there is a module that works really well only for preliminary feedback but does not make sense for graded feedback 🤔

playground/src/components/view_mode/evaluation_mode/conduct_experiment/module_experiment_progress.tsx
FelixTJDietrich: Do it in a follow-up. I think the evaluation mode still works without problems but only for the graded feedback suggestions.

FelixTJDietrich:
Why is the prospector linting check failing? Can we fix this in a follow-up, I will create an issue.
```
Run echo "$HOME/.local/bin" >> $GITHUB_PATH
Creating virtualenv athena-modularized-tH5MAedG-py3.11 in /home/runner/.cache/pypoetry/virtualenvs
Installing dependencies from lock file

pyproject.toml changed significantly since poetry.lock was last generated. Run `poetry lock [--no-update]` to fix the lock file.
```

FelixTJDietrich (approved):
Code LGTM, works as expected on the test server. I will merge it now.

undernagruzez:
> Why is the prospector linting check failing? Can we fix this in a follow-up, I will create an issue.
> 
> ```
> Run echo "$HOME/.local/bin" >> $GITHUB_PATH
> Creating virtualenv athena-modularized-tH5MAedG-py3.11 in /home/runner/.cache/pypoetry/virtualenvs
> Installing dependencies from lock file
> 
> pyproject.toml changed significantly since poetry.lock was last generated. Run `poetry lock [--no-update]` to fix the lock file.
> ```

This is because we use different versions of langchain for different llm modules. We cannot fix it now but this issue will disappear when we decouple modules from the top-level module(I will create a PR by tomorrow).

PR #292, opened 2024-06-09
Add zero shot approach for programming exercises

FelixTJDietrich:
### Motivation and Context

Models exist with 128k context size and more

### Description

Just put everything in context and generate the guided feedback in one shot.
Both graded feedback and guided feedback have two types of requests to LLM: one in guided package, and one in basic_by_file package.

### Steps for Testing

Not possible ATM, playground is broken :(


FelixTJDietrich:
Will just close this PR for now

PR #310, opened 2024-07-06
Text LLM Module: Upgrade to Langchain 0.2.

EneaGore:
### Motivation and Context

Updating to langchain 0.2 . Fixes to comply to pydantic 2 and moving to openai ^1.30. 
Trying to keep it consistent as much as possible with [PR280 from Dmytro](https://github.com/ls1intum/Athena/pull/280) to start working on the shared llm package.
### Steps for Testing

Most changes are in the Module Text LLM package so testing feedback generation there is the main focus.


module_text_llm/pyproject.toml
```diff
@@ -8,14 +8,14 @@ license = "MIT"
 [tool.poetry.dependencies]
 python = "3.11.*"
 athena = {path = "../athena", develop = true}
-openai = "^0.27.8"
-langchain = "^0.0.350"
+openai = "^1.30.5"
+langchain = "^0.2.0"
```
FelixTJDietrich: Can we have `0.2.18` here?

module_text_llm/pyproject.toml
```diff
@@ -8,14 +8,14 @@ license = "MIT"
...
+langchain-openai = "^0.1.7"
```
FelixTJDietrich: This one can be `0.1.16`

module_text_llm/module_text_llm/helpers/models/replicate.py
```diff
@@ -1,8 +1,8 @@
 import os
 from pydantic import Field, PositiveInt
 from enum import Enum
-
-from langchain.llms import Replicate
+from langchain_community.llms import Replicate
```
FelixTJDietrich: Why is  `langchain-community` not in the dependencies?

module_text_llm/pyproject.toml
```diff
@@ -8,14 +8,14 @@ license = "MIT"
 [tool.poetry.dependencies]
 python = "3.11.*"
 athena = {path = "../athena", develop = true}
-openai = "^0.27.8"
-langchain = "^0.0.350"
+openai = "^1.30.5"
```
FelixTJDietrich: openai `1.0.0` supports now explicit client instantiation. 

We can potentially remove the monkey patching I did for getting OpenAI and Azure OpenAI work in parallel.

module_text_llm/module_text_llm/helpers/models/openai.py
```diff
@@ -5,10 +5,14 @@
...
 
 import openai
-from langchain.chat_models import AzureChatOpenAI, ChatOpenAI
-from langchain.llms import AzureOpenAI, OpenAI
-from langchain.llms.openai import BaseOpenAI
+# from langchain.chat_models import AzureChatOpenAI, ChatOpenAI
+# from langchain.llms import AzureOpenAI, OpenAI
+# from langchain.llms.openai import BaseOpenAI
+# from langchain.base_language import BaseLanguageModel
+from langchain_community.llms import OpenAI
```
FelixTJDietrich: This seems wrong? Can't we use `langchain_openai` instead of `langchain_community`?

module_text_llm/module_text_llm/helpers/models/openai.py
```diff
@@ -5,10 +5,14 @@
...
+from langchain_community.llms.openai import BaseOpenAI
```
FelixTJDietrich: Same here

module_text_llm/module_text_llm/helpers/models/openai.py
```diff
@@ -5,10 +5,14 @@
...
 from langchain.base_language import BaseLanguageModel
```
FelixTJDietrich, line 6: Shouldn't this be from `langchain_core.language_models`

module_text_llm/module_text_llm/helpers/models/openai.py
FelixTJDietrich: `from langchain_openai import OpenAI`?

module_text_llm/module_text_llm/helpers/models/openai.py
```diff
@@ -141,47 +145,44 @@ def _openai_client(use_azure_api: bool, is_preference: bool):
     # API client is setup correctly
     yield
```
FelixTJDietrich: Can you check if the above is still necessary? We might be able to remove the monkey patching

module_text_llm/module_text_llm/helpers/llm_utils.py
```diff
@@ -143,20 +150,38 @@ async def predict_and_parse(
         tags.append(f"module-configuration-{experiment.module_configuration_id}")
     if experiment.run_id is not None:
         tags.append(f"run-{experiment.run_id}")
-
-    if supports_function_calling(model):
-        chain = create_structured_output_chain(pydantic_object, llm=model, prompt=chat_prompt, tags=tags)
         
+    chat_prompt.tags = tags
+    
+    if supports_function_calling(model):
```
FelixTJDietrich: Let's not use `supports_function_calling`, I think we want to use `with_structured_output`:
https://python.langchain.com/v0.2/docs/integrations/chat/

Read through, maybe they also have some other resources:
https://python.langchain.com/v0.2/docs/how_to/structured_output/

Also:
https://python.langchain.com/v0.2/docs/how_to/structured_output/#prompting-and-parsing-model-outputs-directly
"Not all models support .with_structured_output(), since not all models have tool calling or JSON mode support. For such models you'll need to directly prompt the model to use a specific format, and use an output parser to extract the structured response from the raw model output."

module_text_llm/module_text_llm/helpers/llm_utils.py
```diff
@@ -143,20 +150,38 @@ async def predict_and_parse(
...
+    
     try:
-        return await chain.arun(**prompt_input)
+        output_dict = await runnable.ainvoke(prompt_input)
+        return pydantic_object.parse_obj(output_dict)
+        #return await chain.arun(**prompt_input)
     except (OutputParserException, ValidationError):
         # In the future, we should probably have some recovery mechanism here (i.e. fix the output with another prompt)
-        return None
+        return None
```
FelixTJDietrich, line 171: Try to simplify this with the new API

module_text_llm/module_text_llm/helpers/models/replicate.py
FelixTJDietrich: But we can also remove Replicate at this stage already IMO. I think you've understood why I did the config the way I did it so the model can be configured via the playground.

module_text_llm/module_text_llm/helpers/models/openai.py
```diff
@@ -207,49 +207,20 @@ def _get_available_models(openai_models: Dict[str, List[str]],
...
-        "curie",
-        "babbage",
-        "ada",
-    ]
-}
-available_deployments = _get_available_deployments(openai_models, _model_aliases)
-available_models = _get_available_models(openai_models, available_deployments)
-
+available_deployments = _get_available_deployments()
+available_models = _get_available_models(available_deployments)
```
FelixTJDietrich, line 110: Why did you remove the `openai_models`? Can you still select the models in the playground for the non-azure openai models? That was the idea 🤔

FelixTJDietrich (changes requested):

FelixTJDietrich:
Did you test with the other modules on the playground? Maybe you need to stack this PR on the Update Playground PR 🤔 
Since the test-server is connected to ma-schwind, Max noticed that the modeling llm module was not working in Artemis with your PR deployed.

PR #323, opened 2024-07-31
`Module_text_llm`: Upgrade dependencies and openai model retrieval

EneaGore:
> [!NOTE]  
> ENVIRONMENT for text llm has been changed.

### Motivation and Context

Major updates for [langchain ](https://python.langchain.com/v0.2/docs/versions/v0_2/)and openai.

### Description

This PR updates the dependencies of the module_text_llm to the latest version. Namely openai 1.37.1 and langcahin 0.2.11


module_text_llm/pyproject.toml
```diff
@@ -8,14 +8,14 @@ license = "MIT"
...
+openai = "1.37.1"
+langchain = "0.2.11"
+langchain-openai = "0.1.19"
 python-dotenv = "^1.0.0"
 nltk = "^3.8.1"
 gitpython = "^3.1.41"
-replicate = "^0.11.0"
-tiktoken = "^0.4.0"
-langsmith = "^0.0.63"
+tiktoken = "^0.7.0"
```
FelixTJDietrich: While you're at it can you remove the carets `^`? I think this would make it more stable for the future.

module_text_llm/pyproject.toml
```diff
@@ -8,14 +8,14 @@ license = "MIT"
 [tool.poetry.dependencies]
 python = "3.11.*"
 athena = {path = "../athena", develop = true}
-openai = "^0.27.8"
-langchain = "^0.0.350"
+openai = "1.37.1"
+langchain = "0.2.11"
+langchain-openai = "0.1.19"
```
FelixTJDietrich: Why is this not a `0.2` version? Can you group the `langchain` dependencies together?

module_text_llm/module_text_llm/helpers/models/openai.py
```diff
@@ -307,7 +152,8 @@ def get_model(self) -> BaseLanguageModel:
                 BaseLanguageModel: The model.
             """
             model = available_models[self.model_name.value]
-            kwargs = model._lc_kwargs
+            kwargs = model.__dict__ #BaseLanguageModel type
+            # kw = model._lc_kwargs
```
FelixTJDietrich: Remove unused comment

module_text_llm/module_text_llm/helpers/models/openai.py
```diff
@@ -1,254 +1,99 @@
...
-
-    # API client is setup correctly
-    yield
-
-
-def _get_available_deployments(openai_models: Dict[str, List[str]], model_aliases: Dict[str, str]):
+if azure_openai_available:
+    os.environ["AZURE_OPENAI_ENDPOINT"]=os.environ["LLM_AZURE_OPENAI_API_BASE"]
+    os.environ["AZURE_OPENAI_API_KEY"]=os.environ["LLM_AZURE_OPENAI_API_KEY"]
+    os.environ["OPENAI_API_VERSION"]=os.environ["LLM_AZURE_OPENAI_API_VERSION"]
```
FelixTJDietrich: Can you update the `.env.example` environment? I thin we are using a newer azure API version. Do you think that it makes sense to set this in the code and not via the environment?

module_text_llm/module_text_llm/helpers/models/openai.py
```diff
@@ -1,254 +1,99 @@
...
 from langchain.base_language import BaseLanguageModel
-
+from langchain_openai import AzureChatOpenAI, AzureOpenAI, ChatOpenAI
 from athena.logger import logger
 from .model_config import ModelConfig
 
 
 OPENAI_PREFIX = "openai_"
 AZURE_OPENAI_PREFIX = "azure_openai_"
+############################################################################################
```
FelixTJDietrich: You can remove this part and just directly define the env variables in the environment now:
`AZURE_OPENAI_API_KEY` no longer collides with `OPENAI_API_KEY`. We can remove the prefix :) 🎉

module_text_llm/module_text_llm/helpers/models/openai.py
```diff
@@ -1,254 +1,99 @@
...
-def _get_available_deployments(openai_models: Dict[str, List[str]], model_aliases: Dict[str, str]):
+if azure_openai_available:
+    os.environ["AZURE_OPENAI_ENDPOINT"]=os.environ["LLM_AZURE_OPENAI_API_BASE"]
+    os.environ["AZURE_OPENAI_API_KEY"]=os.environ["LLM_AZURE_OPENAI_API_KEY"]
+    os.environ["OPENAI_API_VERSION"]=os.environ["LLM_AZURE_OPENAI_API_VERSION"]
+#########################################################################################
+# END Set Enviorment variables that are automatically used by ChatOpenAI/ChatAzureOpenAI#
+#########################################################################################
+
+actually_deployed_azure= [
```
FelixTJDietrich: Do not hardcode this, use the sdk,  you can have two arrays. `azure_models` and `openai_models` and then you fill them with the available models via the API

module_text_llm/module_text_llm/helpers/models/openai.py
```diff
@@ -1,254 +1,99 @@
...
+    "gpt-35-turbo",
+    "gpt-4-turbo",
+    "gpt-4-vision",
+    "gpt-4o"
+]
+
+new_openai_models = []
+    
+def _get_available_deployments():
     available_deployments: Dict[str, Dict[str, Any]] = {
```
FelixTJDietrich: We can do this in the future when it is necessary. For now we are only interested in the `chat_completions`. Once we need embedding models we can add it then and there but for now let's just focus on the code that is necessary :)

module_text_llm/module_text_llm/helpers/models/openai.py
```diff
@@ -1,254 +1,99 @@
...
+actually_deployed_azure= [
+    "gpt-35-turbo",
+    "gpt-4-turbo",
+    "gpt-4-vision",
+    "gpt-4o"
+]
+
+new_openai_models = []
+    
+def _get_available_deployments():
```
FelixTJDietrich: This should get the available azure deployments from the API

module_text_llm/module_text_llm/helpers/models/openai.py
```diff
@@ -1,254 +1,99 @@
...
+        for model in openai.models.list():
+            if model.owned_by == "openai":
+                new_openai_models.append(model.id)
 
     return available_deployments
 
 
-def _get_available_models(openai_models: Dict[str, List[str]], 
-                          available_deployments: Dict[str, Dict[str, Any]]):
+def _get_available_models(available_deployments: Dict[str, Dict[str, Any]]):
```
FelixTJDietrich: This should get the available non-azure models, maybe you can refactor the code a bit in general so it is clearer. You can change it how you like and don't need to necessarily encapsulate in a function, I did it to hide variables.

FelixTJDietrich (changes requested):
Looks good in the playground:
image

Here are some comments. The important part is getting the deployments dynamically. After you're done I can also do a follow-up PR to do some refactoring, I also have some ideas how to can be more maintainable for the future

module_text_llm/pyproject.toml
EneaGore: that is the highest atm

module_text_llm/module_text_llm/helpers/models/openai.py
EneaGore: moved to dependencies, but still using one array for all available models for simplicity

module_text_llm/module_text_llm/helpers/models/openai.py
EneaGore: not possible at the moment to get correct azure deployments from api

module_text_llm/.env.example
```diff
@@ -23,13 +23,14 @@ LLM_EVALUATION_MODEL="azure_openai_gpt-4"
...
 
 # Azure OpenAI [leave blank if not used]
 # Model names prefixed with `azure_openai_` followed by the deployment id, e.g. `azure_openai_gpt-35`
-LLM_AZURE_OPENAI_API_KEY="xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"
-LLM_AZURE_OPENAI_API_BASE="https://ase-eu01.openai.azure.com/" # change base if needed
-LLM_AZURE_OPENAI_API_VERSION="2023-07-01-preview" # change base if needed
+AZURE_OPENAI_API_KEY="xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"
+AZURE_OPENAI_ENDPOINT="https://ase-eu01.openai.azure.com/" # change base if needed
+OPENAI_API_VERSION="2023-07-01-preview" # change base if needed
+AZURE_DEPLOYMENTS= '["gpt-35-turbo","gpt-4-turbo","gpt-4-vision","gpt-4o"]'
```
FelixTJDietrich:
```suggestion
AZURE_DEPLOYMENTS=gpt-35-turbo,gpt-4-turbo,gpt-4-vision,gpt-4o
```
I think a comma separated list would work better for the env

module_text_llm/module_text_llm/helpers/models/openai.py
```diff
@@ -1,254 +1,45 @@
...
+    openai.api_type= "openai"
+    for model in openai.models.list():
+        if model.owned_by == "openai":
+            available_models[OPENAI_PREFIX + model.id] = ChatOpenAI(#type:ignore
+            model=model.id,
+        )
+
+if azure_openai_available:
+    # Will be replaced in the future
+    for deployment in json.loads(os.environ['AZURE_DEPLOYMENTS']):
```
FelixTJDietrich:
```suggestion
    for deployment in os.environ['AZURE_DEPLOYMENTS'].split(','):
```
See comment above

module_text_llm/module_text_llm/helpers/models/openai.py
```diff
@@ -1,254 +1,45 @@
...
+                available_models[AZURE_OPENAI_PREFIX + deployment] = AzureChatOpenAI(deployment_name=deployment, temperature=0, client="")
+  
 if available_models:
     logger.info("Available openai models: %s", ", ".join(available_models.keys()))
 
     OpenAIModel = Enum('OpenAIModel', {name: name for name in available_models})  # type: ignore
-
-
-    default_model_name = "gpt-3.5-turbo"
+    default_model_name = AZURE_OPENAI_PREFIX+"gpt-35-turbo"
```
FelixTJDietrich: I think the `AZURE_OPENAI_PREFIX` is wrong here. This is about the models hosted via non-azure OpenAI?

module_text_llm/.env.example
```diff
@@ -23,13 +23,14 @@ LLM_EVALUATION_MODEL="azure_openai_gpt-4"
...
+OPENAI_API_KEY="xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"
 
 # Azure OpenAI [leave blank if not used]
 # Model names prefixed with `azure_openai_` followed by the deployment id, e.g. `azure_openai_gpt-35`
-LLM_AZURE_OPENAI_API_KEY="xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"
-LLM_AZURE_OPENAI_API_BASE="https://ase-eu01.openai.azure.com/" # change base if needed
-LLM_AZURE_OPENAI_API_VERSION="2023-07-01-preview" # change base if needed
+AZURE_OPENAI_API_KEY="xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"
+AZURE_OPENAI_ENDPOINT="https://ase-eu01.openai.azure.com/" # change base if needed
+OPENAI_API_VERSION="2023-07-01-preview" # change base if needed
```
FelixTJDietrich, line 32: Is this the current version we are using? Does it need updating?

module_text_llm/module_text_llm/helpers/models/openai.py
```diff
@@ -1,254 +1,45 @@
...
-from langchain.chat_models import AzureChatOpenAI, ChatOpenAI
-from langchain.llms import AzureOpenAI, OpenAI
-from langchain.llms.openai import BaseOpenAI
 from langchain.base_language import BaseLanguageModel
-
+from langchain_openai import AzureChatOpenAI, AzureOpenAI, ChatOpenAI
 from athena.logger import logger
 from .model_config import ModelConfig
-
+import json
```
FelixTJDietrich: Can be removed if we change to `split(',')`

module_text_llm/module_text_llm/helpers/models/openai.py
```diff
@@ -1,254 +1,45 @@
...
 from pydantic import Field, validator, PositiveInt
 from enum import Enum
-
 import openai
-from langchain.chat_models import AzureChatOpenAI, ChatOpenAI
-from langchain.llms import AzureOpenAI, OpenAI
-from langchain.llms.openai import BaseOpenAI
 from langchain.base_language import BaseLanguageModel
-
+from langchain_openai import AzureChatOpenAI, AzureOpenAI, ChatOpenAI
```
FelixTJDietrich: Are we even using `AzureOpenAI`? Maybe you have some unused imports.

FelixTJDietrich (changes requested):
Good job so far :) 
I will test the functionality soon

module_text_llm/.env.example
EneaGore: These ones are the new names.

module_text_llm/module_text_llm/helpers/models/openai.py
EneaGore: I thought we used the same config for both, at least so it appeared to me.

module_text_llm/module_text_llm/helpers/models/openai.py
```diff
@@ -1,254 +1,44 @@
...
+                available_models[AZURE_OPENAI_PREFIX + deployment] = AzureChatOpenAI(deployment_name=deployment, temperature=0, client="")
+  
 if available_models:
     logger.info("Available openai models: %s", ", ".join(available_models.keys()))
 
     OpenAIModel = Enum('OpenAIModel', {name: name for name in available_models})  # type: ignore
-
-
-    default_model_name = "gpt-3.5-turbo"
+    default_model_name = AZURE_OPENAI_PREFIX+"gpt-35-turbo"
```
FelixTJDietrich:
```suggestion
    default_model_name = "gpt-35-turbo"
```
This is just the default model in the case that we have nothing configured as `LLM_DEFAULT_MODEL` for the non-azure case.

FelixTJDietrich (changes requested):

FelixTJDietrich:
Other than the one thing this PR is ready to merge, I also tested it on the playground and it works as expected. I verified it over my LangSmith instance and the right models were called.

module_text_llm/module_text_llm/helpers/models/openai.py
```diff
@@ -1,254 +1,44 @@
...
+                available_models[AZURE_OPENAI_PREFIX + deployment] = AzureChatOpenAI(deployment_name=deployment, temperature=0, client="")
+  
 if available_models:
     logger.info("Available openai models: %s", ", ".join(available_models.keys()))
 
     OpenAIModel = Enum('OpenAIModel', {name: name for name in available_models})  # type: ignore
-
-
-    default_model_name = "gpt-3.5-turbo"
+    default_model_name = "gpt-35-turbo"
```
FelixTJDietrich:
```suggestion
    default_model_name = None
```

FelixTJDietrich (changes requested):

FelixTJDietrich (approved):
LGTM

PR #326, opened 2024-08-05
`module_modeling_LLM`: Fix the handling of BPMN diagrams without pools

LeonWehrhahn:
### Motivation and Context
The change is required to fix an issue in the BPMNSerializer where BPMN diagrams without pools were not being serialized. This caused the diagrams to not be added correctly to the prompt, resulting in wrong feedback on BPMN diagram submissions.

### Description
Modified the BPMNSerializer to handle elements without pools by assigning them to a default unnamed pool.

### Steps for Testing
Use the Playground to:
- Test with the "Pizza Delivery" example, which initially had issues due to the lack of pools, and ensure that all elements are correctly serialized and the LLM provides relevant feedback.
- Use the "Obtaining a Loan" example, which includes pools, to confirm it still works as expected.


maximiliansoelch (approved):
Tested on TS1, feedback is now also generated for models without a pool.
Code looks good 👍

PR #327, opened 2024-08-05
Fix OpenAI model configuration

FelixTJDietrich:
### Motivation and Context

While deploying wen noticed that hardcoding the deployments was not ideal. Also `gpt-4o` was not selectable for OpenAI.

### Description

- Fix OpenAI models
- Remove hardcoded Azure deployments
- Use old Azure API to fetch deployments, once it breaks we will migrate

### Steps for Testing

Check if TextLLM still works in the playground.

### Screenshots

image


FelixTJDietrich:
Checked on TS1 and it works. I also changed the env on the TS1

PR #346, opened 2024-10-11
Move llm_core

FelixTJDietrich:
### Motivation and Context

We only have assessment modules in `/modules`

### Description

 Move `llm_core` on the same level as `athena`

### Steps for Testing


FelixTJDietrich:
Tested on TS2 works as expected

PR #349, opened 2024-10-22
Add ollama support to llm_core

EneaGore:
### Motivation and Context

https://github.com/ls1intum/Athena/pull/403

### Description

Adds support for Ollama models in Athena.
Implements a Chain of Thought approach with prompt chaining.
### Steps for Testing

Depoy locally (with vpn) or directly to TS2.
Choose the ollama CoT approach (You can also select simply the model, however it might not return structured output).


EneaGore:
Refer to https://github.com/ls1intum/Athena/pull/403 instead

PR #350, opened 2024-10-24
Multiple Selectable Approaches for Text (Chain of Thought with Prompt Chaining)

EneaGore:
### Motivation and Context

This PR provides a new structure for different approaches within the text llm module. It already adds a new approach "chain of thought".
### Description

The standard approach has been moved in a folder named basic_approach. Each approach has its own folder so they can be easily added, or deleted in case it is not needed anymore. The approach controller and config need to be adjusted when a new approach is added or removed.
### Steps for Testing

1. Select text llm module
2. Observe the new approach selector, pick either one.
3. Confirm that the request is processed and that the correct one was called (either use langsmith locally or see that, for example, in the case of chain of thought with chain prompting there are 2 llm calls in the meta data output) 


llm_core/llm_core/models/__init__.py
```diff
@@ -6,6 +6,7 @@
 
 
 DefaultModelConfig: Type[ModelConfig]
+MiniModelConfig: ModelConfig
```
FelixTJDietrich, line 9: Why have `MiniModel` config? We just want one model config per provider, i.e. (Azure)OpenAI and Ollama

llm_core/llm_core/utils/llm_utils.py
```diff
@@ -75,7 +75,7 @@ def supports_function_calling(model: BaseLanguageModel):
     Returns:
         boolean: True if the model supports function calling, False otherwise
     """
-    return isinstance(model, ChatOpenAI)
+    return isinstance(model, ChatOpenAI)  or isinstance(model, AzureChatOpenAI)
```
FelixTJDietrich, line 78: But now it won't work for OpenAI without Azure?

llm_core/llm_core/utils/predict_and_parse.py
```diff
@@ -36,13 +36,14 @@ async def predict_and_parse(
     if experiment.run_id is not None:
         tags.append(f"run-{experiment.run_id}")
 
-    structured_output_llm = model.with_structured_output(pydantic_object, method="json_mode")
-    chain = RunnableSequence(
-        chat_prompt,
-        structured_output_llm
-    )
+    structured_output_llm = model.with_structured_output(pydantic_object)
```
FelixTJDietrich: Are we aiming to use function calling or json mode? Do you know if something changed on the Langchain side regarding json mode?

Function calling still has the problem of raising errors when the format is unexpected

llm_core/llm_core/utils/predict_and_parse.py
```diff
@@ -36,13 +36,14 @@ async def predict_and_parse(
...
+    # chain = RunnableSequence(
+    #     chat_prompt,
+    #     structured_output_llm
+    # )
+    chain = chat_prompt | structured_output_llm
 
     try:
-        return await chain.ainvoke(prompt_input, config={"tags": tags})
+        return await chain.ainvoke(prompt_input, config={"tags": tags}) # type: ignore #
```
FelixTJDietrich: Why do you need the type ignore, can we handle it some other way?

modules/text/module_text_llm/module_text_llm/approaches/approach_config.py
```diff
@@ -0,0 +1,16 @@
...
+from enum import Enum
+
+class ApproachType(str, Enum):
+    basic = "BasicApproach"
+    chain_of_thought = "ChainOfThought"
+    
+class ApproachConfig(BaseModel, ABC):
+    max_input_tokens: int = Field(default=3000, description="Maximum number of tokens in the input prompt.")
+    model: ModelConfigType = Field(default=DefaultModelConfig())  # type: ignore
+    type: ApproachType = Field(..., description="The type of approach config")
```
FelixTJDietrich: Does it work with the playground? 
If so, this seems like an elegant solution :)

modules/text/module_text_llm/module_text_llm/approaches/basic_approach/config.py
```diff
@@ -0,0 +1,26 @@
+from module_text_llm.approaches.approach_config import ApproachConfig
+from pydantic import Field, BaseModel
+from typing import Literal
+
+
+from module_text_llm.approaches.basic_approach.prompts.generate_suggestions import (
```
FelixTJDietrich: I think we should just have a config directory with an `__init__.py` for the approach config i.e. `BasicApproachConfig`. Then separate files for prompts etc. i.e. `generate.py` with the `GenerateSuggestionsPrompt` and also the contents of `module_text_llm.approaches.basic_approach.prompts.generate_suggestions`

modules/text/module_text_llm/module_text_llm/approaches/basic_approach/prompts/generate_suggestions.py
```diff
@@ -17,10 +17,14 @@
 # Grading instructions
 {grading_instructions}
 Max points: {max_points}, bonus points: {bonus_points}\
+    
+    Respomd in json format.
```
FelixTJDietrich: Is the typo intentional? Do you know if it improves performance for function calling to have this there? Maybe we put this as part of the format prompt function.

modules/text/module_text_llm/module_text_llm/approaches/chain_of_thought_approach/config.py
```diff
@@ -0,0 +1,36 @@
...
+)
+
+from module_text_llm.approaches.chain_of_thought_approach.prompts.refined_cot_suggestions import (
+  system_message as generate_refined_cot_suggestions_system_message, 
+  human_message as generate_refined_cot_suggestions_human_message
+)
+
+from module_text_llm.approaches.approach_config import ApproachConfig
+
+class CoTGenerateSuggestionsPrompt(BaseModel):
```
FelixTJDietrich: I think you don't have to prefix it with CoT if we move it in a `config/` directory.

modules/text/module_text_llm/module_text_llm/approaches/approach_config.py
```diff
@@ -0,0 +1,16 @@
...
+from llm_core.models import ModelConfigType, DefaultModelConfig
+from enum import Enum
+
+class ApproachType(str, Enum):
+    basic = "BasicApproach"
+    chain_of_thought = "ChainOfThought"
+    
+class ApproachConfig(BaseModel, ABC):
+    max_input_tokens: int = Field(default=3000, description="Maximum number of tokens in the input prompt.")
+    model: ModelConfigType = Field(default=DefaultModelConfig())  # type: ignore
```
FelixTJDietrich: I think `max_input_tokens` and `model` is approach specific. We can keep it like this for now but an approach can use multiple different models in the future.

modules/text/module_text_llm/module_text_llm/approaches/chain_of_thought_approach/generate_suggestions.py
```diff
@@ -0,0 +1,152 @@
...
+class InitialAssessment(BaseModel):
+    title: str = Field(description="Very short title, i.e. feedback category or similar", example="Logic Error")
+    description: str = Field(description="Feedback description")
+    line_start: Optional[int] = Field(description="Referenced line number start, or empty if unreferenced")
+    line_end: Optional[int] = Field(description="Referenced line number end, or empty if unreferenced")
+    credits: float = Field(0.0, description="Number of points received/deducted")
+    reasoning: str = Field(description="Reasoning why the feedback was given")
+    impprovment_suggestion: str = Field(description="Suggestion for improvement for the student")
+
+class InitialAssessmentModel(BaseModel):
```
FelixTJDietrich: Since they are kind of part of the prompt we could think of moving them in a `config/generate/` directory. 🤔

llm_core/llm_core/models/__init__.py
EneaGore: I was thinking that if i use multiple prompt chains the gpt-4o mini could be better suited and to set it as the default in the approach instead of the actual default which costs significantly more i.d.r

llm_core/llm_core/utils/llm_utils.py
EneaGore: it should work for both. The problem was that before it was doing the extra steps in chat prompt creation that are required for replicate (and will be for ollama) but which are not for azure and openai

modules/text/module_text_llm/module_text_llm/approaches/approach_config.py
EneaGore: it does :D

modules/text/module_text_llm/module_text_llm/approaches/basic_approach/prompts/generate_suggestions.py
EneaGore: No the type wasn't intentional . Function calling required that the prompt contained the word json in it.

llm_core/llm_core/utils/predict_and_parse.py
undernagruzez: This is addressed in my [pr
](https://github.com/ls1intum/Athena/pull/342/files#diff-59de72932727318da8304170c5fa30f582552d7be1fcda9ed7243273261e98e0)

FelixTJDietrich (approved):
Structure and code looks good now. 

We will have to fix the playground issue in a follow-up

PR #375, opened 2024-12-09
Use the default model for the text Chain of Thought approach

EneaGore:
### Motivation and Context

The chain of thought approach was using the open ai api 4o-mini as default, which would cause an error if only the azure key was provided.
### Description

The CoT approach now uses the default model.
### Steps for Testing


maximiliansoelch (approved):
Tested locally and the text_llm module is starting again

PR #385, opened 2024-12-31
Implement strategy Pattern for approach selection and use dynamic approach discovery

EneaGore:
### Motivation and Context

Nested if-else statements are considered an anti-pattern as they reduce code readability and maintainability. This change refactors the code to implement the strategy pattern, ensuring the suggestion generator calls the appropriate method dynamically.

Additionally, previously adding a new approach required modifications across three classes, which increased coupling and complexity. By adopting a dynamic approach discovery mechanism, adding a new approach now requires changes to only a single class, thereby improving modularity and reducing coupling.
### Description

This pull request introduces several key changes to enhance the flexibility and extensibility of the configuration management system by integrating the Strategy Pattern and dynamically handling approach configurations.

- Strategy Pattern Integration: The SuggestionStrategyFactory class has been introduced to manage the creation and retrieval of suggestion strategies based on dynamically discovered configuration classes. By implementing the Strategy Pattern, different strategies can be easily selected and applied based on the configuration provided, allowing for modular and customizable behavior without modifying existing code.

- Dynamic Approach Discovery: The Configuration class now dynamically loads all approach configurations available in the module_text_llm package using the discover_approach_configs function. This eliminates the need for manually updating a union of approach classes, simplifying the addition of new approaches.
### Steps for Testing

Deploy to a test server or test locally.
Switch between approaches and request suggestions (make sure that the model chosen is appropriate for text, sometimes the default model in the UI is set to the first model)
Confirm that the correct approach was selected and no errors occur.


modules/text/module_text_llm/module_text_llm/chain_of_thought_approach/generate_suggestions.py
```diff
@@ -19,7 +18,7 @@
 from module_text_llm.chain_of_thought_approach.prompt_generate_feedback import AssessmentModel
 
 
-async def generate_suggestions(exercise: Exercise, submission: Submission, config: ChainOfThoughtConfig, debug: bool) -> List[Feedback]:
+async def generate_suggestions(exercise: Exercise, submission: Submission, config: ApproachConfig, debug: bool) -> List[Feedback]:
```
FelixTJDietrich: Why can't this be `ChainOfThoughtConfig`? I think using `ApproachConfig` is too generic, how can you access the config of the approach then?

modules/text/module_text_llm/module_text_llm/approach_controller.py
```diff
@@ -1,15 +1,103 @@
...
+            config (object): The configuration object for which the strategy is required.
+
+        Returns:
+            object: An instance of the strategy class associated with the given configuration.
+
+        Raises:
+            ValueError: If no strategy is found for the given configuration type.
+        """
+        if not SuggestionStrategyFactory._strategies:
+            SuggestionStrategyFactory.initialize_strategies()
```
FelixTJDietrich: Maybe you should just do this on the top level when the module loads? But idk, this works

modules/text/module_text_llm/module_text_llm/approach_controller.py
```diff
@@ -1,15 +1,103 @@
...
+                module = importlib.import_module(full_module_name)
+                for name, obj in inspect.getmembers(module, inspect.isclass):
+                    if base_class is None or (issubclass(obj, base_class) and obj is not base_class):
+                        classes[name] = obj
+
+    recursive_import(base_package)
+    return classes
+
+
+class SuggestionStrategyFactory:
```
FelixTJDietrich: How are the strategies listed for the playground? We still need this union type. Also, is it possible to move it to the athena package instead of keeping it specific to the text module. We can also refactor it in a second step, but this would also be nice for other modules.

modules/text/module_text_llm/module_text_llm/approach_config.py
```diff
@@ -1,16 +1,17 @@
...
-class ApproachType(str, Enum):
-    basic = "BasicApproach"
-    chain_of_thought = "ChainOfThought"
+from abc import ABC, abstractmethod
+from athena.text import Exercise, Submission
     
 class ApproachConfig(BaseModel, ABC):
     max_input_tokens: int = Field(default=3000, description="Maximum number of tokens in the input prompt.")
     model: ModelConfigType = Field(default=DefaultModelConfig())
     type: str = Field(..., description="The type of approach config")
```
FelixTJDietrich, line 9: If you are anyways doing discovering stuff dynamically and using the class name etc. you could also just skip this field and use the class's name 🤔

modules/text/module_text_llm/module_text_llm/approach_controller.py
FelixTJDietrich: Keeping it for every module would also make it easier when dealing with the playground implementation. I think we'd need to update the config schema endpoint to respond with all discovered strategy configs.

FelixTJDietrich (changes requested):

modules/text/module_text_llm/module_text_llm/chain_of_thought_approach/generate_suggestions.py
EneaGore: It still works the same because of Polymorphism, I had to use the generic ApproachConfig to avoid a circular import, when the ChainofThoughtConfig is passed it works the same.

modules/text/module_text_llm/module_text_llm/approach_config.py
FelixTJDietrich: it's fine as is

modules/text/module_text_llm/module_text_llm/basic_approach/generate_suggestions.py
```diff
@@ -10,11 +11,11 @@
 )
 from llm_core.utils.predict_and_parse import predict_and_parse
 
-from module_text_llm.config import BasicApproachConfig
+# from module_text_llm.config import BasicApproachConfig
 from module_text_llm.helpers.utils import add_sentence_numbers, get_index_range_from_line_range, format_grading_instructions
 from module_text_llm.basic_approach.prompt_generate_suggestions import AssessmentModel
 
-async def generate_suggestions(exercise: Exercise, submission: Submission, config: BasicApproachConfig, debug: bool, is_graded: bool) -> List[Feedback]:
+async def generate_suggestions(exercise: Exercise, submission: Submission, config:ApproachConfig, debug: bool, is_graded: bool) -> List[Feedback]:
```
FelixTJDietrich:
```suggestion
async def generate_suggestions(exercise: Exercise, submission: Submission, config: ApproachConfig, debug: bool, is_graded: bool) -> List[Feedback]:
```
space

modules/text/module_text_llm/module_text_llm/create_approach_script.py
FelixTJDietrich, line 1: Cool script, maybe document its existence in the future. But I think some students might find it. I think we can keep it as is

FelixTJDietrich (dismissed):
I will apply the one suggestion that changes the formatting, the rest looks pretty good

FelixTJDietrich (approved):

PR #403, opened 2025-01-27
Ollama models

EneaGore:
### Motivation and Context

This update integrates Ollama models from the GPU server into Athena's available models. Only the absolute neccessary code to have ollama available and working.
### Description

Adds support for Ollama models in Athena.
### Steps for Testing

Depoy locally (with vpn) or directly to the test servers with the correct credentials on their env.
Choose a simple exercise and it should produce structured output for it with the basic apporach. For example coupling and cohesion usually works.

### DIFF
--- a/llm_core/llm_core/models/ollama.py
+++ b/llm_core/llm_core/models/ollama.py
@@ -71,7 +71,7 @@ if os.environ.get('GPU_USER') and os.environ.get('GPU_PASSWORD') and os.environ.
                     return v
                 
                 def get_model(self) -> BaseLanguageModel:
-                    print("Getting Model: ", self.model_name.value)
+                    logger.info("Getting Model: ", self.model_name.value)
                     """Get the model from the configuration.
 
                     Returns:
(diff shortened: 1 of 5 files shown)


llm_core/llm_core/models/ollama.py
```diff
@@ -0,0 +1,116 @@
...
+                    model = name,
+                    base_url = os.environ["OLLAMA_ENDPOINT"],
+                    headers = auth_header,
+                    format = "json"
+                ) for name in ollama_models
+            } 
+
+        default_model_name = "llama3.3:latest"
+        LlamaModel = Enum('LlamaModel', {name: name for name in available_models}) # type: ignore
+        class OllamaModelConfig(ModelConfig):
```
FelixTJDietrich, line 42: Since the descriptions are getting rendered in the Playground they could be more descriptive but this can also be a follow-up

FelixTJDietrich (approved):
Great work! 
Works as expected on the playground:

image

image

PR #406, opened 2025-02-26
Self consistency

EneaGore:
### Motivation and Context

This change introduces a new self-consistency approach that leverages an **ensemble of approaches**. The implementation runs multiple approaches concurrently and then uses a weighted scoring formula, taking into account credits and structured grading instructions, to select the most self-consistent result. This method addresses issues with inconsistent feedback from previous implementations.

### Description
This PR implements the following enhancements:
- **Ensemble Execution:** Executes multiple approaches concurrently to generate feedback.
- **Weighted Scoring:** Aggregates feedback using a weighted formula based on credits and structured grading criteria.
- **Selection Mechanism:** Automatically selects and returns the most self-consistent approach based on computed scores.

These changes aim to improve the robustness and reliability of feedback generation in the Text LLM Module.

### Steps for Testing
1. **Select Text LLM Module:** Open the module within the application.
2. **Choose Self Consistency Approach:** Select the Self Consistency approach from the available options.
3. **Provide Input:** Choose a text exercise and submit a corresponding submission.
4. **Verify Output:** Confirm that feedback is generated successfully and that the most self-consistent approach is selected based on the weighted scoring.


FelixTJDietrich (approved):
Code LGTM, I will verify it on TS2 and merge it if it works :) 
Thanks for your work and having it in a mergable state already!

